PR #8, opened 2020-12-21
[llvm] Add support for .TEXT section size reward

ChrisCummins:
The IrInstructionCount reward space only counts the number of LLVM-IR
instructions in the current module. The actual size of the compiled
code may differ from this.

This adds a new NativeTextSizeBytes reward space which returns the
size of the .text section of the compiled module.


ChrisCummins:
Superseded by #9.

PR #11, opened 2020-12-22
Add microbenchmark suite.

broune:
Add a microbenchmark suite.

I also fixed a bug with "make test" where it would fail to provide a file that has to exist or else make test will hang forever. I can back it out if one instead prefers to fix it by changing random_search to not require the presence of that file.

### DIFF
--- a/tests/benchmarks/bench_test.py
+++ b/tests/benchmarks/bench_test.py
@@ -1,0 +1,80 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Microbenchmarks for CompilerGym environments."""
+import gym
+import pytest
+
+from compiler_gym.envs import CompilerEnv, LlvmEnv
+from compiler_gym.service import CompilerGymServiceConnection
+from tests.envs.llvm.fixtures import SERVICE_BIN
+from tests.test_main import main
+
+pytest_plugins = ["tests.envs.llvm.fixtures"]
+
+# Redefine this fixture to run only the local LLVM env.
+@pytest.fixture(scope="function")
+def env() -> CompilerEnv:
+    """Create an LLVM environment."""
+    env = gym.make("llvm-v0")
+    env.require_dataset("cBench-v0")
+    try:
+        yield env
+    finally:
+        env.close()
+
+
+# Redefine this fixture since running all of the benchmarks in cBench would
+# take too long, but we do want to use at least one small and one large
+# benchmark to see both per-invocation overhead and overhead that is a result
+# of the size of the fixture.
+#
+# adpcm is small and jpeg-d is large. ghostscript is the largest but that
+# one takes too long.
+@pytest.fixture(params=["cBench-v0/adpcm", "cBench-v0/jpeg-d"])
+def benchmark_name(request) -> str:
+    yield request.param
+
+
+# The observation benchmark is too slow for a large input.
+@pytest.fixture(params=["cBench-v0/adpcm"])
+def fast_benchmark_name(request) -> str:
+    yield request.param
+
+
+def test_make_local(benchmark):
+    benchmark(lambda: gym.make("llvm-v0").close())
+
+
+def test_make_service(benchmark):
+    service = CompilerGymServiceConnection(SERVICE_BIN)
+    try:
+        benchmark(lambda: LlvmEnv(service=service.connection.url).close())
+    finally:
+        service.close()
+
+
+def test_reset(benchmark, env: CompilerEnv, benchmark_name):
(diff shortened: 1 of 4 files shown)


tests/benchmarks/bench_test.py
```diff
@@ -0,0 +1,60 @@
...
+      service.close()
+
+def test_reset(benchmark, env: CompilerEnv, benchmark_name):
+    benchmark(env.reset, benchmark_name)
+
+def test_step(benchmark, env: CompilerEnv, benchmark_name):
+    env.reset(benchmark_name)
+    benchmark(env.step, 0)
+
+def test_observation(benchmark, env: CompilerEnv, fast_benchmark_name):
```
ChrisCummins: Could you add an [observation_space](https://github.com/facebookresearch/CompilerGym/blob/development/tests/envs/llvm/fixtures.py#L49-L51) so that we can benchmark each observation space? They take different amounts of time to compute, it would be great to track those differences.
```suggestion
def test_observation(benchmark, env: CompilerEnv, fast_benchmark_name, observation_space):
```

tests/benchmarks/bench_test.py
```diff
@@ -0,0 +1,60 @@
...
+    env.reset(fast_benchmark_name)
+    benchmark(lambda: env.observation["Programl"])
```
ChrisCummins:
```suggestion
    benchmark(lambda: env.observation[observation_space])
```

tests/benchmarks/bench_test.py
```diff
@@ -0,0 +1,60 @@
...
+
+def test_reward(benchmark, env: CompilerEnv, benchmark_name):
```
ChrisCummins: Same again here but for reward spaces. Please parameterize this benchmark by the `reward_space` fixture so that we can track the performance of different reward functions (at the moment we only have instruction-count based rewards but this will change soon when we add support for optimizing for performance).

ChrisCummins (changes requested):
Thanks a lot @broune this looks great! I have a couple of small feature requests that would be nice to get in before merging.

In `tests/benchmarks/bench_test.py`, could you define a new environment fixture and use that rather than `env`? The reason is that the `env` fixture instantiates a couple of different environments with different backend configurations which is important for testing but irrelevant for benchmarking. Something like this would do the trick:

```py
@pytest.fixture(scope="function")
def env() -> CompilerEnv:
    """Create an LLVM environment."""
    env = gym.make("llvm-v0")
    env.require_dataset("cBench-v0")
    try:
        yield env
    finally:
        env.close()
```

Cheers,
Chris

ChrisCummins:
Oh and one final thing - please run the code formatter on the new python file so that the pre-commit checks doesn't complain :)

The easiest way to do that is using [pre-commit](https://github.com/facebookresearch/CompilerGym/blob/development/.pre-commit-config.yaml):

```sh
brew install pre-commit buildifier llvm prototool python
pre-commit install  # install the pre-commit hooks
git commit --amend  # will run a bunch of pre-commit hooks to format the sources
```

tests/benchmarks/bench_test.py
broune: Done.

tests/benchmarks/bench_test.py
broune: Done.

tests/benchmarks/bench_test.py
broune: Done.

broune:
I set it up to override env so that it only tries the local version as requested. If we do not want to benchmarking through the service, should I then also remove test_make_service?

I set up the precommit on my end. There is still a precommit error, but it is in README.md that I'm not modifying - it wasn't me!

Removing the service targets removed a bunch of benchmarks, though adding in reward and observation spaces added more. Here's a command line for running with Bazel and its output on my machine:

  https://pastebin.com/raw/tC9Gmfpc

ChrisCummins:
Hi @broune,

> I set it up to override env so that it only tries the local version as requested. If we do not want to benchmarking through the service, should I then also remove test_make_service?

Great, thanks! I like that we benchmark the local/service inits, it's interesting that they take a meaningfully different amount of time.

> I set up the precommit on my end. There is still a precommit error, but it is in README.md that I'm not modifying - it wasn't me!

Ah no worries then :-) If there's every a false-positive that you would like to ignore, you can use the `git commit --no-verify` flag to disable pre-commit formatters.

> Removing the service targets removed a bunch of benchmarks, though adding in reward and observation spaces added more. Here's a command line for running with Bazel and its output on my machine:
> 
> https://pastebin.com/raw/tC9Gmfpc

Brilliant, thanks a lot for the pastebin, those numbers will be super useful when making back-of-the-envelope estimates of computation costs, and for measuring perf improvements!

LGTM, merging.

Cheers,
Chris

PR #36, opened 2021-01-11
Add a `make itest` target.

ChrisCummins:
This is equivalent to running `make test` on change. Useful for
iterative development.


hughleat (approved):
LGTM

PR #28, opened 2020-12-29
Add compiler gym actor_critic example.

broune:
Add compiler gym actor-critic example.

This example is simplified by only training for a single benchmark at a time, using a subset of 15 LLVM compiler passes and using a fixed episode length.

Here is a paste with a bunch of data from just playing around with the parameters:

  https://pastebin.com/raw/HgS1GCEe

If run with no parameters, it takes 11 seconds to find a good configuration on my macbook. Likely this could be made much faster but this is just supposed to be a fairly simple example, so I think that is fine.

Example output from running python actor_critic.py :

Seed: 101
Episode length: 5
Exploration: 0.00%
Learning rate: 0.008
Action space: ['-simplifycfg', '-reg2mem', '-early-cse-memssa', '-gvn-hoist', '-gvn', '-instsimplify', '-instcombine', '-jump-threading', '-loop-extract', '-loop-reduce', '-loop-versioning', '-newgvn', '-mem2reg', '-sroa', '-structurizecfg']
Episode 1	Last reward: 52.00	Avg reward: 52.00	Best reward: 52.00	Last loss: 0.410147	Avg loss: 0.410147	
Episode 100	Last reward: 97.00	Avg reward: 118.10	Best reward: 153.00	Last loss: -0.045142	Avg loss: 0.432208	
Episode 200	Last reward: 140.00	Avg reward: 130.32	Best reward: 153.00	Last loss: 3.115863	Avg loss: 0.219633	
Episode 300	Last reward: 131.00	Avg reward: 135.27	Best reward: 153.00	Last loss: -0.087667	Avg loss: -0.016292	
Episode 400	Last reward: 143.00	Avg reward: 140.41	Best reward: 153.00	Last loss: -0.556506	Avg loss: 0.031324	
Episode 500	Last reward: 152.00	Avg reward: 145.52	Best reward: 153.00	Last loss: 0.899966	Avg loss: 0.032300	
Episode 600	Last reward: 152.00	Avg reward: 150.30	Best reward: 153.00	Last loss: 0.076278	Avg loss: -0.001032	
Episode 700	Last reward: 152.00	Avg reward: 151.88	Best reward: 153.00	Last loss: -0.059516	Avg loss: -0.024967	
Episode 800	Last reward: 153.00	Avg reward: 152.25	Best reward: 153.00	Last loss: -0.138806	Avg loss: -0.066543	
[...]

PR #37, opened 2021-01-11
Re-merge actor critic and add a smoke test

ChrisCummins:
I'm not sure what happened here, but the #28 PR merge seems to have gotten lost!

This PR cherry picks @broune's original #28 changes, and adds a simple smoke test to catch regressions from API changes.


broune (approved):
Looks good to me.

Issue #22, opened 2020-12-28
Github CI actions are flaky because of network errors during bazel fetch


ChrisCummins:
## 🐛 Bug

When running bazel on GitHub, the repository fetch stage occasionally fails with error `Tried to reconnect at offset xxx but server didn't support it`. Example failure:

```
ERROR: An error occurred during the fetch of repository 'com_github_grpc_grpc':
   java.io.IOException: Error downloading [https://github.com/grpc/grpc/archive/v1.34.0.tar.gz] to /private/var/tmp/_bazel_runner/079b33da0ef26a5d53e8a115deeb0011/external/com_github_grpc_grpc/v1.34.0.tar.gz: Tried to reconnect at offset 393,216 but server didn't support it
ERROR: no such package '@com_github_grpc_grpc//bazel': java.io.IOException: Error downloading [https://github.com/grpc/grpc/archive/v1.34.0.tar.gz] to /private/var/tmp/_bazel_runner/079b33da0ef26a5d53e8a115deeb0011/external/com_github_grpc_grpc/v1.34.0.tar.gz: Tried to reconnect at offset 393,216 but server didn't support it
```

## To Reproduce

1. Trigger a [CI](https://github.com/facebookresearch/CompilerGym/actions?query=workflow%3ACI) github action.

ChrisCummins:
A second error type that is possibly related is a failure to load `@local_config_cc`:

```
bazel  build --config=ci //:package
2021/01/20 05:10:34 Downloading https://releases.bazel.build/3.1.0/release/bazel-3.1.0-darwin-x86_64...
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
ERROR: /DEFAULT.WORKSPACE.SUFFIX:317:1: error loading package '@local_config_cc//': Unable to load file '@local_config_cc_toolchains//:osx_archs.bzl': file doesn't exist and referenced by '//external:cc_toolchain'
ERROR: /private/var/tmp/_bazel_runner/079b33da0ef26a5d53e8a115deeb0011/external/bazel_tools/tools/cpp/BUILD:64:1: Target '@bazel_tools//tools/cpp:malloc' depends on toolchain '@local_config_cc//:cc-compiler-darwin', which cannot be found: error loading package '@local_config_cc//': Unable to load file '@local_config_cc_toolchains//:osx_archs.bzl': file doesn't exist'
ERROR: While resolving toolchains for target @bazel_tools//tools/cpp:malloc: Analysis failed
[...]

PR #38, opened 2021-01-15
Add examples/explore.py that explores the states that are reachable

broune:
for a certain set of actions within a given episode length.

$ python explore.py --env=llvm-ic-v0 --benchmark=cBench-v0/dijkstra --episode_length=10 --nproc 16 --topn 1000 --show_topn=15

Output: https://pastebin.com/raw/dhyYFZQb

Partial sequences that end up in the same state are deduplicated,
dramatically reducing the number of states to be searched. There is
also an option to use a beam search.

Duplicates are currently detected using a hashcode/fingerprint of the
IR, but it appears that there is some state in an env that is not
captured by printing the IR. Also, and perhaps related, the number of
duplicates identified is slightly different when using parallelism. So
there is some more digging to be done.

Conservatively set the changed flag to true in all LLVM envs, as the
env relies on it being accurate for correctness and it is not
reliable. Some passes say no change occurred when a change did occur.
Let me know if people want this addressed in some other way.

### DIFF
--- a/compiler_gym/envs/llvm/service/LlvmEnvironment.cc
+++ b/compiler_gym/envs/llvm/service/LlvmEnvironment.cc
@@ -161,24 +161,15 @@ Status LlvmEnvironment::takeAction(const ActionRequest& request, ActionReply* re
   RETURN_IF_ERROR(verifyModuleStatus(benchmark().module()));
 
   if (eagerObservationSpace().has_value()) {
-    // Compute new observation if needed.
-    if (!reply->action_had_no_effect()) {
-      eagerObservation_ = {};
-      RETURN_IF_ERROR(getObservation(eagerObservationSpace().value(), &eagerObservation_));
-    }
+    eagerObservation_ = {};
+    RETURN_IF_ERROR(getObservation(eagerObservationSpace().value(), &eagerObservation_));
     *reply->mutable_observation() = eagerObservation_;
   }
 
   if (eagerRewardSpace().has_value()) {
-    if (reply->action_had_no_effect()) {
-      // Action had no effect, so no reward.
-      reply->mutable_reward()->set_reward(0);
-    } else {
-      // Compute new reward if needed.
-      eagerReward_ = {};
-      RETURN_IF_ERROR(getReward(eagerRewardSpace().value(), &eagerReward_));
-      *reply->mutable_reward() = eagerReward_;
-    }
+    eagerReward_ = {};
+    RETURN_IF_ERROR(getReward(eagerRewardSpace().value(), &eagerReward_));
+    *reply->mutable_reward() = eagerReward_;
   }
 
   return Status::OK;
@@ -188,13 +179,7 @@ void LlvmEnvironment::runPass(llvm::Pass* pass, ActionReply* reply) {
   llvm::legacy::PassManager passManager;
   setupPassManager(&passManager, pass);
 
-  bool changed = passManager.run(benchmark().module());
-
-  // TODO: Fix changed. changed is not reliable in LLVM as currently
-  // implemented and the gym environment relies on it being accurate,
-  // so conservatively hardcode it to true for now.
-  changed = true;
-
+  const bool changed = passManager.run(benchmark().module());
   reply->set_action_had_no_effect(!changed);
 }
 
@@ -207,12 +192,6 @@ void LlvmEnvironment::runPass(llvm::FunctionPass* pass, ActionReply* reply) {
     changed |= (passManager.run(function) ? 1 : 0);
   }
   changed |= (passManager.doFinalization() ? 1 : 0);
-
-  // TODO: Fix changed. changed is not reliable in LLVM as currently
-  // implemented and the gym environment relies on it being accurate,
-  // so conservatively hardcode it to true for now.
-  changed = true;
-
   reply->set_action_had_no_effect(!changed);
 }
 
--- a/compiler_gym/envs/llvm/service/LlvmEnvironment.cc
+++ b/compiler_gym/envs/llvm/service/LlvmEnvironment.cc
@@ -188,7 +188,13 @@ void LlvmEnvironment::runPass(llvm::Pass* pass, ActionReply* reply) {
   llvm::legacy::PassManager passManager;
   setupPassManager(&passManager, pass);
 
-  const bool changed = passManager.run(benchmark().module());
+  bool changed = passManager.run(benchmark().module());
+
+  // TODO: Fix changed. changed is not reliable in LLVM as currently
+  // implemented and the gym environment relies on it being accurate,
+  // so conservatively hardcode it to true for now.
+  changed = true;
+
   reply->set_action_had_no_effect(!changed);
 }
 
@@ -201,6 +207,12 @@ void LlvmEnvironment::runPass(llvm::FunctionPass* pass, ActionReply* reply) {
     changed |= (passManager.run(function) ? 1 : 0);
   }
   changed |= (passManager.doFinalization() ? 1 : 0);
+
+  // TODO: Fix changed. changed is not reliable in LLVM as currently
+  // implemented and the gym environment relies on it being accurate,
+  // so conservatively hardcode it to true for now.
+  changed = true;
+
   reply->set_action_had_no_effect(!changed);
 }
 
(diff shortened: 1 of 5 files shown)


examples/explore.py
```diff
@@ -0,0 +1,506 @@
...
+)
+flags.DEFINE_integer("episode_length", 5, "The number of steps in each episode.")
+flags.DEFINE_integer(
+    "nproc", cpu_count(), "The number of parallel worker threads to run."
+)
+flags.DEFINE_integer(
+    "topn",
+    0,
+    "If positive, explore only the top n states "
+    "for each sequence length. This is in effect the widt "
```
ChrisCummins:
```suggestion
    "for each sequence length. This is in effect the width "
```

compiler_gym/envs/llvm/service/LlvmEnvironment.cc
```diff
@@ -188,7 +188,13 @@ void LlvmEnvironment::runPass(llvm::Pass* pass, ActionReply* reply) {
   llvm::legacy::PassManager passManager;
   setupPassManager(&passManager, pass);
 
-  const bool changed = passManager.run(benchmark().module());
+  bool changed = passManager.run(benchmark().module());
+
+  // TODO: Fix changed. changed is not reliable in LLVM as currently
+  // implemented and the gym environment relies on it being accurate,
+  // so conservatively hardcode it to true for now.
+  changed = true;
```
ChrisCummins: I think there's no harm in reporting the LLVM 'changed' value, even if it is unsuitable for our needs. The problem is that the gym environment shouldn't rely on the value. So I would prefer to leave the value of `changed` as it was, but instead to remove the conditional code in `LlvmEnvironment.cc` that uses the `action_had_no_effect` value to skip work.

It would be interesting to see what effect this has on the `step()` microbenchmark time.

examples/explore.py
```diff
@@ -0,0 +1,506 @@
...
+        self._env = env_from_flags(benchmark_from_flags())
+        try:
+            # Project onto the subset of transformations that have
+            # been specified to be used.
+            if not FLAGS.actions:
+                self._action_indices = list(range(len(self._env.action_space.names)))
+            else:
+                self._action_indices = [
+                    self._env.action_space.flags.index(a) for a in FLAGS.actions
+                ]
```
ChrisCummins, line 75: nitpick

```suggestion
            if FLAGS.actions:
                self._action_indices = [
                    self._env.action_space.flags.index(a) for a in FLAGS.actions
                ]
            else:
                self._action_indices = list(range(len(self._env.action_space.names)))
```

ChrisCummins: As a side note it would be great if there was an API like:

```py
env.action_space = env.action_space.make_subset(["-mem2reg", ...])
```

to handle this action index mapping rather than having to define a wrapper class.

examples/explore.py
```diff
@@ -0,0 +1,506 @@
...
+
+
+NO_EDGE = -1
+
+
+class Node:
+    def __init__(self, reward_sum, edge_count):
+        self.reward_sum = reward_sum
+        self.edges = [NO_EDGE] * edge_count
+        self.back_edge = None
```
ChrisCummins, line 124: FYI (no change needed) [networkx](https://networkx.org/documentation/stable/tutorial.html) provides a nice interface for building and working with attributed graphs in Python

examples/explore.py
```diff
@@ -0,0 +1,506 @@
...
+    # TODO: There is some sort of state in the env that is not
+    # captured by this. Figure out what it is and fix it. Also
+    # consider adding a fingerprint observation to env.
+    if False:
+        # BitcodeFile is slower, so using Ir instead.
+        path = env.observation["BitcodeFile"]
+        with open(path, "rb") as f:
+            data = f.read()
+    else:
+        data = env.observation["Ir"].encode()
```
ChrisCummins, line 197:
```suggestion
    data = env.observation["Ir"].encode()
```

The `BitcodeFile` and `Ir` observations call [different LLVM APIs](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/LlvmEnvironment.cc#L209-L230), but I think the approach you used is exactly right for what you want.

Something to keep in mind for later would be to add an observation space that returns the [Benchmark::hash()](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/Benchmark.h#L69) SHA1, rather than serializing the entire IR and doing the checksum from the python side. Not required for this PR.

ChrisCummins (changes requested):
Thanks for this @broune, and great job debugging the inconsistencies in the LLVM env. I ran the script on my end. Couple of observations:

- the depth top-n reports seem to vary across benchmarks (at least on the couple of cBench programs that I tried), would be interesting to compare/characterize state graphs across benchmarks.
- the top-n report quickly saturates with "reward-equivalent" states (e.g. depth 9 in your pastebin). It would be interesting to see what are the differences in the IRs that produce those.

I wouldn't hardcode the `changed` value, I would just not use it in the `LlvmEnvironment`. Couple of other nitpicks left inline. Otherwise looking good!

Cheers,
Chris

examples/explore.py
broune: Done.

compiler_gym/envs/llvm/service/LlvmEnvironment.cc
broune: Done.

Just for posterity, this is the command line I used to run the microbenchmarks:

  bazel test --test_output=all --test_arg=-s :bench_test --nocache_test_results --test_arg=-k='test_step and cBench-v0 and adpcm'

Results here:

  https://pastebin.com/raw/81esBiVw

The difference, if any, is in the noise, which makes sense since I placed some CHECK(false) and determined that eager reward and observation are not turned on for that microbenchmark. One could add a microbenchmark for this, but not clear if it's actionable for this change now.

examples/explore.py
broune: Good idea, it could then be tied to a common flag and picked up in env_from_flags() with some nicer error messages etc. than what I did here. I'll add it to my todo.

examples/explore.py
broune: Good to know, thanks.

examples/explore.py
broune: I think it makes sense leaving the TODO here indicating both that something is not captured, so that this isn't a real fingerprint even though the function is named as though it is (which otherwise could trip someone up the same way it did me for a while), and that having this on the env itself as an observation is preferable. I'd also like to keep both IR/bitcode versions of computing the "fingerprint" since (surprisingly) they can distinguish separate sets of envs and e.g. one thing that would be logical to do in adding a fingerprint observation on the env is to determine whether the set of envs distinguished by the bitcode fingerprint here is identical to that distinguished by the fingerprint on the env. Which it ought to be, but I'm not confident, e.g. I wonder what the interplay here is with the option "ShouldPreserveUseListOrder" used when calculating the hash / bitcode file and any other options. Once this is all cleared up then the entire function should be removed.

broune:
> * the top-n report quickly saturates with "reward-equivalent" states (e.g. depth 9 in your pastebin). It would be interesting to see what are the differences in the IRs that produce those.
> 
Yes, it is especially interesting that globaldce is managing to change the IR without changing the size. I spent some time thinking about measures to promote beam diversity like they do in Adams et al, but it got complicated. Some of this can likely be canonicalized out, but stuff like reordering instructions is a legitimate thing that can likely change follow-on actions, you'd expect there to be some of that and it's not correct to ignore it in the hash for that reason. At least the shortest sequence that achieves a certain level of reward will be more sensible.

examples/explore.py
ChrisCummins: Yep, hooking that into the common flags would be great!

examples/explore.py
ChrisCummins: Sounds good, thanks for the explanation.

ChrisCummins:
> Yes, it is especially interesting that globaldce is managing to change the IR without changing the size.

Hmm, it could be that the pass adds an attribute to the IR which is used by later passes/codegen, but that is just me guessing. Making `BitcodeFile` observations and [llvm-diff](https://llvm.org/docs/CommandGuide/llvm-diff.html) might be useful here.

ChrisCummins:
Thanks for the changes and clarifications @broune. This looks good to me. Noting that the failing test signal is caused by https://github.com/facebookresearch/CompilerGym/issues/22, ready to merge!

Cheers,
Chris

ChrisCummins (approved):

PR #41, opened 2021-01-22
Deallocate episodes when they end

ChrisCummins:
Due to an oversight on my part episodes were never actually deleted from the map of active sessions on a call to EndEpisode().


broune (approved):
Looks good to me!

PR #43, opened 2021-01-22
Add semantics validation for cBench benchmarks

ChrisCummins:
Add semantics validation for cBench benchmarks. This is achieved by adding a new validation callback mechanism that, when invoked, compiles the given cBench benchmark to a binary and executes it using prepared datasets. The output of the program, along with any generated output files, is differential tested against a copy of the program compiled without optimizations. A change in program behavior that is detected by this mechanism is reported.

Calling `compiler_gym.validate_state()` on a benchmark that supports semantics validation will automatically run it.

The core of the implementation is in `compiler_gym/envs/llvm/dataset.py`. It defines a set of library functions so that these validation callbacks can be defined ad-hoc for cBench in quite a succinct form, e.g.:

```py
validator(
    benchmark="benchmark://cBench-v0/ghostscript",
    cmd="$BIN -sDEVICE=ppm -dNOPAUSE -dQUIET -sOutputFile=output.ppm -- 1.ps",
    data=["office_data/1.ps"],
    outs=["output.ppm"],
    linkopts=["-lm", "-lz"],
    pre_execution_callback=setup_ghostscript_library_files,
)
```

As part of #45 we may want to make a public API similar to this and move it into the dataset definitions.

Multiple validation callbacks can be defined for a single benchmark. Where a benchmark matches multiple validators, they are executed in parallel.


WORKSPACE
```diff
@@ -201,11 +201,93 @@ filegroup(
 )
 
 http_archive(
+    name = "ctuning-ai",
+    build_file_content = """
+filegroup(
```
broune: nit: I would have understood this faster if it were indented more than build_file_content is, which would be nice. (unless it doesn't parse with prefix whitespace on each line in the string?)

compiler_gym/envs/compiler_env.py
```diff
@@ -794,8 +795,13 @@ def register_dataset(self, dataset: Dataset) -> None:
             >>> env.benchmark = "my-dataset-v0/1"
 
         :param dataset: A :class:`Dataset` instance describing the new dataset.
+        :return: :code:`True` if the dataset was registered added, else :code:`False`.
```
broune: It says both registered and added.

compiler_gym/envs/llvm/datasets.py
```diff
@@ -110,12 +158,519 @@
...
+                walltime_seconds=timeout_seconds,
+                error=f"Benchmark failed to complete with {timeout_seconds} timeout.",
+            )
+
+        if not ignore_returncode and process.returncode:
+            output = None
+            try:
+                output = stdout.decode("utf-8")
+            except UnicodeDecodeError:
+                pass
```
broune: Shouldn't this be an error?

compiler_gym/envs/llvm/datasets.py
```diff
@@ -110,12 +158,519 @@
...
+            return BenchmarkExecutionResult(
+                walltime_seconds=timer.time,
+                error=f"Benchmark exited with returncode {process.returncode}. Output: {output}",
+            )
+        return BenchmarkExecutionResult(walltime_seconds=timer.time, output=stdout)
+    finally:
+        bitcode_file.unlink()
+
+
+@lru_cache(maxsize=1024)
```
broune: How large are the cache entries in .output? Any chance we'll exhaust available memory?

compiler_gym/envs/llvm/datasets.py
```diff
@@ -110,12 +158,519 @@
...
+
+def _compose_validators(*validators):
+    def composed(env):
+        # Shortcut when there is only a single validator to run.
+        if len(validators):
+            return validators[0](env)
+
+        # Run validators simultaneously in parallel threads.
+        with ThreadPoolExecutor() as executor:
+            futures = [executor.submit(validator, env) for validator in validators]
```
broune: Are these all using the same env and is that safe? If so would be good to have a comment saying why.

compiler_gym/util/temporary_working_directory.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+import tempfile
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Optional, Union
+
+
+@contextmanager
+def temporary_working_directory(directory: Optional[Union[str, Path]] = None) -> Path:
+    """Temporarily set the working directory.
```
broune, line 15: Would be good to describe in a comment how to pop back to the original state, i.e. get the code after "yield" to run that cleans things up. This isn't obvious to me.

tests/envs/compiler_env_test.py
```diff
@@ -47,5 +51,56 @@ def test_state_from_csv_invalid_format():
     assert str(ctx.value).startswith("Failed to parse input: `abcdef`: ")
 
 
+@pytest.fixture(scope="function")
+def env() -> CompilerEnv:
```
broune, line 55: This doesn't seem to be called?

compiler_gym/envs/llvm/datasets.py
```diff
@@ -110,12 +158,519 @@
...
+            validate_result=validate_result,
+            linkopts=linkopts,
+            os_env=env,
+            pre_execution_callback=pre_execution_callback,
+        )
+    )
+    return True
+
+
+def _compose_validators(*validators):
```
broune: This seems not to be called?

compiler_gym/envs/llvm/datasets.py
```diff
@@ -110,12 +158,519 @@
...
+    validate_result: Optional[
+        Callable[[BenchmarkExecutionResult], Optional[str]]
+    ] = None,
+    pre_execution_callback: Optional[Callable[[], None]] = None,
+):
+    """Generate a validation callback for a cBench benchmark."""
+    input_files = input_files or []
+    output_files = output_files or []
+
+    def validator(env):
```
broune: Would it make sense to ensure that the benchmarks satisfy the LLVM sanitizers like asan, msan etc. both before and after optimization? (if so, can just be a TODO, but those seem valuable - if the original benchmark doesn't satisfy a sanitizer, then optimizations that mess it up aren't necessarily wrong, while introducing san* failures would be bad.)

broune (approved):
Looks good to me, had a few questions.

WORKSPACE
ChrisCummins: Good point. I can't indent the inline `build_file_content` because Starlark is whitespace-sensitive, but I can pull the definition out into another file.

compiler_gym/envs/compiler_env.py
ChrisCummins: Thanks!

compiler_gym/envs/llvm/datasets.py
ChrisCummins: 👍

compiler_gym/envs/llvm/datasets.py
ChrisCummins: Oo good question. In practice most benchmarks produce a few lines of output at most so this shouldn't be a problem, but there isn't a limit. I've added a 10k char truncation just to bound the max size of this cache at ~10MB.

compiler_gym/envs/llvm/datasets.py
ChrisCummins: The validation callback doesn't modify the environment, this is safe. Added a comment.

compiler_gym/util/temporary_working_directory.py
ChrisCummins: Done

tests/envs/compiler_env_test.py
ChrisCummins: This declares a [pytest fixture](https://docs.pytest.org/en/stable/fixture.html#back-to-fixtures) so that tests in this file that take an `env` argument will use this.

compiler_gym/envs/llvm/datasets.py
ChrisCummins: Called in `get_llvm_benchmark_validation_callback()`, but it wasn't clear so I've refactored.

compiler_gym/envs/llvm/datasets.py
ChrisCummins: Update, I was dead wrong about this! There was a bug with multiple threads racing to set a different working directory. I fixed the bug and added a couple of new tests.

compiler_gym/envs/llvm/datasets.py
ChrisCummins: I really like this idea! Yep let's stick a TODO for this. It will take a bit of extra plumbing to get it working (we don't currently ship the clang rt libs) but the value is clear

ChrisCummins:
Thanks for the review @broune . This definitely isn't the most elegant code, but the idea will be to turn this proof-of-concept validation into a public API once we've ironed out a few more of the details.

Cheers,
Chris

compiler_gym/envs/llvm/datasets.py
ChrisCummins: Opened #49 to follow up on this.

PR #48, opened 2021-02-03
Add an LlvmEnv.fork() method

ChrisCummins:
Calling `env.fork()` creates a duplicate environment with the current                                                                                                                                                                     
state. For long episodes, this may be more efficient than resetting a                                                                                                                                                                   
fresh state and replaying the episode.       

Example usage:

```py
>>> env = gym.make("llvm-v0")
# ... use env
>>> new_env = env.fork()
>>> new_env.actions == env.actions
True
```                                                                                                                                                                                           
                                                                                                                                                                                                                                        
This implementation is totally ad-hoc for LLVM. It creates a new                                                                                                                                                                        
compiler service and makes a temporary benchmark out of the current                                                                                                                                                                     
environment state by serializing to and reading from a bitcode file. A                                                                                                                                                                  
lower overhead `fork()` could be implemented on the C++ service side by                                                                                                                                                                   
creating a new `LlvmEnvironment` for the existing service, but that                                                                                                                                                                       
would require the addition of a new API call. Let's see if `fork()`                                                                                                                                                                       
performance becomes a key factor before pursuing that.

### DIFF
--- a/compiler_gym/envs/compiler_env.py
+++ b/compiler_gym/envs/compiler_env.py
@@ -417,13 +417,7 @@ class CompilerEnv(gym.Env):
             self._user_specified_benchmark_uri = benchmark
         elif isinstance(benchmark, Benchmark):
             self._user_specified_benchmark_uri = benchmark.uri
-            # Register the custom benchmark, and record the Benchmark object
-            # in case of environment restart.
-            self._custom_benchmarks[benchmark.uri] = benchmark
-            self.service(
-                self.service.stub.AddBenchmark,
-                AddBenchmarkRequest(benchmark=[benchmark]),
-            )
+            self._add_custom_benchmarks([benchmark])
         else:
             raise TypeError(f"Unsupported benchmark type: {type(benchmark).__name__}")
 
@@ -575,11 +569,8 @@ class CompilerEnv(gym.Env):
             self.service = CompilerGymServiceConnection(
                 self.service_endpoint, self.connection_settings
             )
-            # Re-register any custom benchmarks.
-            self.service(
-                self.service.stub.AddBenchmark,
-                AddBenchmarkRequest(benchmark=list(self._custom_benchmarks.values())),
-            )
+            # Re-register the custom benchmarks with the new service.
+            self._add_custom_benchmarks(self._custom_benchmarks.values())
 
         self.action_space_name = action_space or self.action_space_name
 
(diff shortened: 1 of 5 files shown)


ChrisCummins:
FYI the test runners are on fire at the moment, please ignore red signals :)

image

compiler_gym/envs/compiler_env.py
```diff
@@ -193,8 +193,8 @@ def __init__(
         self,
         service: Union[str, Path],
         benchmark: Optional[Union[str, Benchmark]] = None,
-        observation_space: Optional[str] = None,
-        reward_space: Optional[str] = None,
+        observation_space: Optional[Union[str, ObservationSpaceSpec]] = None,
+        reward_space: Optional[Union[str, RewardSpaceSpec]] = None,
```
hughleat, line 204: Do the comments below for these parameters not need to be updated? They aren't entirely clear.
Also, for the string version, maybe the comment should explain from where to get the strings?
How come action_space doesn't also have a spec?

compiler_gym/envs/compiler_env.py
```diff
@@ -466,19 +473,30 @@ def observation_space(self) -> Optional[ObservationSpaceSpec]:
...
+        observation_space_name = (
+            observation_space.id
+            if isinstance(observation_space, ObservationSpaceSpec)
+            else observation_space
+        )
         if self.in_episode:
             warnings.warn(
                 "Changing eager observation space has no effect until reset() is called."
             )
+
```
hughleat, line 506: This code is very similar `reward_space` (line 430), but some small differences.
Why isn't there:
  `self._eager_observation_space: str = observation_space_name or ""`
Like there is:
  `self._eager_reward_space: str = reward_space_name or ""`
Feels like there might be some inconsistency in typing?

compiler_gym/envs/compiler_env.py
hughleat: Are `ObservationSpaceSpec` and `RewardSpaceSpec ` sealed? Can the user make their own objects?

tests/envs/llvm/fork_env_test.py
```diff
@@ -0,0 +1,93 @@
...
+        new_env.reset()
+        assert ir(env) == ir(new_env)
+    finally:
+        new_env.close()
+
+
+FUZZ_TIME_SECONDS = 5
+
+
+def test_fork_state_fuzz_test(env: LlvmEnv):
```
hughleat, line 75: Maybe also check that steps in `new_env` lead to the same state as a step in `env`?
i.e.
- do rand actions to env
- new_env = env.fork
- for a in a_few_rand_actions
    - new_env.step(a)
    - env.step(a)
    - check new_env == env

hughleat (approved):
LGTM.

compiler_gym/envs/compiler_env.py
ChrisCummins: Good catch, I've updated the docstrings. A user can make their own observation spaces using [add_derived_space()](https://facebookresearch.github.io/CompilerGym/compiler_gym/views.html#compiler_gym.views.ObservationView.add_derived_space).

Action spaces don't have a spec because at the moment they are simply `gym.spaces.Discrete` categorical spaces. Once we add support for other space types that will be updated

compiler_gym/envs/compiler_env.py
ChrisCummins: Yes, great catch! Fixed

tests/envs/llvm/fork_env_test.py
ChrisCummins: Good idea, done.

ChrisCummins:
Thanks for the review @hughleat!

PR #62, opened 2021-02-09
Temporarily disable flaky cBench benchmarks

ChrisCummins:
Each of these benchmarks has been found to have issues when used for
difftesting, possibly due to user error, possibly due to misconfigured
code. Further investigation is needed. This reduces the number of
cBench benchmarks whose semantics can be validated to 11.


ChrisCummins:
Merged into #67.

broune (commented):
Looks good to me.

PR #63, opened 2021-02-09
Prune some irrelevant passes from the LLVM action space.

ChrisCummins:
Deselect some passes that are obviously irrelevant (e.g. Objective-C
refcounting passes), used for debugging (e.g. BarrierNoop), or passes
that require profiling data as
input (e.g. PostInlineEntryexitinstrumenter).

### DIFF
--- a/compiler_gym/envs/llvm/service/passes/config.py
+++ b/compiler_gym/envs/llvm/service/passes/config.py
@@ -157,9 +157,15 @@ CREATE_PASS_NAME_MAP = {
 
 # A list of pass names that should be excluded from the action space.
 _EXCLUDED_PASSES = {
-    # Garbage collection.
+    # Irrelevant garbage collection passes.
     "StripGCRelocates",
     "PlaceBackedgeSafepointsImpl",
+    "PlaceSafepointsPass",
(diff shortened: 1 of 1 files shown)


compiler_gym/envs/llvm/service/passes/config.py
```diff
@@ -173,6 +179,9 @@
     "DAH",  # Bugpoint only.
     "MetaRenamerPass",
     "PAEvalPass",
+    "BarrierNoop",  # Used for debugging pass manager.
+    "StripNonLineTableDebugInfoPass",  # Debug stripping.
+    "StripDeadDebugInfoPass",  # Debug stripping.
```
broune, line 184: Do we always know that the input binary has no debug info to strip?

broune (commented):
Looks good to me!

compiler_gym/envs/llvm/service/passes/config.py
ChrisCummins: Good question! Debug info is stripped as part of the [environment setup](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/LlvmEnvironment.cc#L137-L144). I copied this behavior from `opt`

PR #67, opened 2021-02-10
Service and validation fixups.

ChrisCummins:
* Fixes a bug where `env.benchmark` reports no value until `env.reset()` is called.
* Increases the timeouts on service startup.
* Makes runfiles_path() use paths relative to the repo root
* Fixes a bug in error message generation for difftesting.


hughleat (approved):
LGTM

PR #69, opened 2021-02-11
Move reward calculation into Python frontend

ChrisCummins:
Instead of calculating reward in the service backend, a new `Reward`
class is added to the python frontend and used to compute rewards
based on observation values generated by the service.

## Why?

The goal of this change is twofold:

1. to simplify the backend service API by merging the
  GetObservation(), GetReward(), and TakeAction() operators into a
  single Step() call. This in turn enables more efficient usage as
  multiple calls can be replaced by a single Step() invocation.
2. to make it easier for users to define their own reward
  spaces. Like with observation spaces, learning techniques are
  sensitive to the reward space design decisions and so experimenting
  with reward shaping/scaling/design is a common use case. By
  implementing rewards in Python that simplifies the experimentation.

## How?

The new RPC `Step()` interface accepts a list of actions and a list of
observations. The service first applies the actions and then generates
the observations. The frontend python code then uses those
observations to compute a reward value.

One downside of this approach is that it means it is no longer to add
support for a new compiler without writing any Python, but this seems
like a reasonable tradeoff for the benefits.

## Who is affected?

There is no public API change; 99% of users won't notice a difference. Folks who want to define their own reward space can now do so without recompiling the backend.

PR #74, opened 2021-02-12
Generate enum declarations for LLVM capabilities.

### LINKED ISSUE #72: Make all strings accessible from objects.
## 🚀 Feature

String use is error prone and has little support from IDEs.
Instead of writing:
`    env.observation_space = "IRInstructionCountOs"`
which has an error that an IDE can't help me with because that space does not exist,
Write:
`    env.observation_space = compiler_gym.llvm.observation_spaces.IrInstructionCountOz`
Or something like that.
This way the IDE or debugger will auto complete for me and may even tell me there is an error before running it.

`compiler_gym.llvm.observation_spaces.IrInstructionCountOz` can still be a string, so minimal changes are required.

## Motivation

I don't like remembering things or continually having to look things up when the computer can do it for me.

The classes for this could be auto generated from the c++?

## Pitch

See above

## Alternatives

Don't do it?

## Additional context

None

ChrisCummins:
This adds new enums `compiler_gym.envs.llvm.observation_spaces` and
`compiler_gym.envs.llvm.reward_spaces` that define the string names of
the available LLVM observation and reward spaces.

In https://github.com/facebookresearch/CompilerGym/issues/72#issuecomment-778216106 I suggested adding this feature to `compiler_gym.bin.service`. This turned out to be a bit tricky without introducing a circular dependency. Instead, I made a small ad-hoc script to generate the enums.

Fixes #72.

### DIFF
--- a/compiler_gym/envs/llvm/BUILD
+++ b/compiler_gym/envs/llvm/BUILD
@@ -5,7 +5,10 @@
 
 py_library(
     name = "llvm",
-    srcs = ["__init__.py"],
+    srcs = [
+        "__init__.py",
+        ":specs.py",
+    ],
     data = [
         "//compiler_gym/envs/llvm/service:compiler_gym-llvm-service",
     ],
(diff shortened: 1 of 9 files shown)


ChrisCummins:
Blocking this until #69 is merged.

PR #77, opened 2021-02-17
Runs random walk in gym env and remove redundant actions

JD-ETH:
Fixes #{issue number}
Use reward and environment identical to actor critic example without having to locally compile the environment. Adapted rewards as well. 
Example output: 
```

=== Step 100 ===
Action:       -consthoist (changed=False)
Reward:       0.0
Step time:    763.2us

Completed 100 steps in 2.098s (47.7 steps / sec).
Total reward: -14.0
Max reward:   75.0 (-535.71% at step 47)
Effective actions from trajectory: -loop-extract-single, -gvn-hoist, -lcssa, -gvn, -lcssa, -loop-extract-single, -dce, -loop-extract, -structurizecfg
```

Random walk typically increase lr count rather then reducing them it seems.

### DIFF
--- a/examples/actor_critic.py
+++ b/examples/actor_critic.py
@@ -120,7 +120,7 @@ class CustomEnv:
         self._env = gym.make("llvm-v0", reward_space=FLAGS.reward)
         try:
             self._env.require_dataset("cBench-v0")
-            self._env.reset(benchmark="benchmark://"+FLAGS.benchmark)
+            self._env.reset(benchmark=FLAGS.benchmark)
 
             # Project onto the subset of transformations that have
             # been specified to be used.
--- a/examples/actor_critic.py
+++ b/examples/actor_critic.py
@@ -120,7 +120,7 @@ class CustomEnv:
         self._env = gym.make("llvm-v0", reward_space=FLAGS.reward)
         try:
             self._env.require_dataset("cBench-v0")
-            self._env.reset(benchmark="benchmark://" + FLAGS.benchmark)
+            self._env.reset(benchmark=FLAGS.benchmark)
 
             # Project onto the subset of transformations that have
             # been specified to be used.
--- a/examples/actor_critic.py
+++ b/examples/actor_critic.py
@@ -120,7 +120,7 @@ class CustomEnv:
         self._env = gym.make("llvm-v0", reward_space=FLAGS.reward)
         try:
             self._env.require_dataset("cBench-v0")
-            self._env.reset(benchmark="benchmark://"+FLAGS.benchmark)
+            self._env.reset(benchmark="benchmark://" + FLAGS.benchmark)
 
             # Project onto the subset of transformations that have
             # been specified to be used.
--- a/examples/actor_critic.py
+++ b/examples/actor_critic.py
@@ -120,7 +120,7 @@ class CustomEnv:
         self._env = gym.make("llvm-v0", reward_space=FLAGS.reward)
         try:
             self._env.require_dataset("cBench-v0")
-            self._env.reset(benchmark=FLAGS.benchmark)
+            self._env.reset(benchmark="benchmark://"+FLAGS.benchmark)
 
             # Project onto the subset of transformations that have
             # been specified to be used.
--- a/examples/random_walk.py
+++ b/examples/random_walk.py
@@ -10,8 +10,9 @@ Example usage:
     $ python3 examples/random_walk.py --env=llvm-v0 --step_min=100 --step_max=100 \
         --benchmark=cBench-v0/dijkstra --reward=IrInstructionCount
 """
-import random
 import hashlib
+import random
+
 import humanize
 from absl import app, flags
 
@@ -42,8 +43,10 @@ def run_random_walk(env: CompilerEnv, step_count: int) -> None:
 
     "Use hashes of the environment to record its internal state"
     hashes, actions, rewards = [], [], []
+
     def encode_env_state(env):
         return hashlib.sha1(env.ir.encode("utf-8")).hexdigest()
+
     step_num = 0
     with Timer() as episode_time:
         env.reset()
@@ -87,7 +90,9 @@ def run_random_walk(env: CompilerEnv, step_count: int) -> None:
 
     def minimize_action_sequence(hashes, actions):
         """First pass removes actions that did not change the internal states"""
-        actions = [actions[i] for i in range(len(actions)) if hashes[i] != hashes[i+1]]
+        actions = [
+            actions[i] for i in range(len(actions)) if hashes[i] != hashes[i + 1]
+        ]
         """Second pass removes cancelling actions in the trajectory"""
         cancelling_dict = {"mem2reg": "reg2mem"}
         for act, counteract in cancelling_dict.items():
--- a/examples/random_walk.py
+++ b/examples/random_walk.py
@@ -4,23 +4,11 @@
 # LICENSE file in the root directory of this source tree.
 """Perform a random walk of the action space of a CompilerGym environment.
 
-This program launches a CompilerGym service and runs a random number of steps, at
-each step selecting a random action, and recording the observations and rewards
-to stdout.
-
 Example usage:
 
-    # Build the CompilerGym service binary that you want to use.
-    $ bazel build -c opt //examples/example_compiler_gym_service
-
-    # Run a random walk.
-    $ bazel run -c opt //compiler_gym/bin:random_walk -- \
-        --local_service_binary=$PWD/bazel-bin/examples/example_compiler_gym_service/service/service \
-        --program=foo \
-        --observation=features \
-        --reward=codesize \
-        --step_min=100 \
-        --step_max=100
+    # Run a random walk on cBench example program using instruction count reward.
+    $ python3 examples/random_walk.py --env=llvm-v0 --step_min=100 --step_max=100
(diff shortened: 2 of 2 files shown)


examples/actor_critic.py
```diff
@@ -120,7 +120,7 @@ def __init__(self):
         self._env = gym.make("llvm-v0", reward_space=FLAGS.reward)
         try:
             self._env.require_dataset("cBench-v0")
-            self._env.reset(benchmark=FLAGS.benchmark)
+            self._env.reset(benchmark="benchmark://"+FLAGS.benchmark)
```
ChrisCummins: I don't think this should be necessary

examples/random_walk.py
```diff
@@ -4,23 +4,11 @@
...
-    $ bazel run -c opt //compiler_gym/bin:random_walk -- \
-        --local_service_binary=$PWD/bazel-bin/examples/example_compiler_gym_service/service/service \
-        --program=foo \
-        --observation=features \
-        --reward=codesize \
-        --step_min=100 \
-        --step_max=100
+    # Run a random walk on cBench example program using instruction count reward.
+    $ python3 examples/random_walk.py --env=llvm-v0 --step_min=100 --step_max=100
+    --benchmark=benchmark://cBench-v0/dijkstra --reward=IrInstructionCount
```
ChrisCummins:
```suggestion
    # Run a random walk on cBench example program using instruction count reward.
    $ python3 examples/random_walk.py --env=llvm-v0 --step_min=100 --step_max=100 \
        --benchmark=benchmark://cBench-v0/dijkstra --reward=IrInstructionCount
```

super nitpick :)

examples/random_walk.py
```diff
@@ -77,22 +65,25 @@ def run_random_walk(env: CompilerEnv, step_count: int) -> List[float]:
...
+    print(f"Total reward: {sum(rewards)}")
     print(
-        f"Max reward:   {max(rewards)} ({reward_delta(max(rewards))} "
+        f"Max reward:   {max(rewards)} ({reward_percentage(max(rewards), rewards)} "
         f"at step {humanize.intcomma(rewards.index(max(rewards)) + 1)})"
     )
-    return rewards
-
+    def remove_no_change(rewards, actions):
+        return [a for (r, a) in zip(rewards, actions) if r != 0]
```
ChrisCummins, line 84: I like the idea of reporting "effective actions", but we should be careful about how "effective" is defined. It would be quite possible for an action to have an effect on the environment that does not yield a reward, but changes the state in a way that meaningfully affects the environment. E.g. if the reward space is instruction count, consider an action that both removes 10 instructions and then adds 10 instructions. It has certainly affected the environment, but will produce a reward of zero.

If the goal of `remove_no_change()` is to filter the list of actions but still produce the same final state, that will be a little bit more tricky, but would be extremely useful.

One thing that might be helpful for this is the `action_had_no_effect` element that is returned in the `info` dict of all LLVM environments:

```py
>>> observation, reward, done, info = env.step(action)
>>> info
{'action_had_no_effect': True, 'new_action_space': False}
```

Although worth noting that this value is produced by LLVM and it actually under-reports changes, i.e. it is possible for an action to modify the state but for LLVM to report that it did nothing. For the LLVM environments, the most-complete view of the state can be obtained by hashing the LLVM-IR of the environment:

```py
>>> import hashlib
>>> hashlib.sha1(env.ir.encode("utf-8")).hexdigest()
'24c6dd4a6d4cb8dfb818577070432f41d4cd78b1'
```

Tracking how that hash changes over the course of a trajectory may well provide a more accurate view on which actions had an effect than the reward values.

Consider also that there are potentially subsequences of actions that effectively cancel each other out, so those too could be pruned. For example, the LLVM actions `-mem2reg` and `-reg2mem` have opposite effects:

```
$ python -m compiler_gym.bin.manual_env --env=llvm-ic-v0 --benchmark=cBench-v0/dijkstra
Initialized environment in 111.6ms
Welcome to the CompilerGym Shell!
---------------------------------
Type help or ? for more information.
The 'tutorial' command will give a step by step guide.

compilergym:cBench-v0/dijkstra> reset
Reset in 85.2ms

compilergym:cBench-v0/dijkstra> action -mem2reg
Action -mem2reg
Reward: 0.625806
Actions -mem2reg in 1.1ms with reward 0.6258064516129033.

compilergym:cBench-v0/dijkstra> action -reg2mem
Action -reg2mem
Reward: -0.703226
Actions -reg2mem in 909.6us with reward -0.7032258064516129.

compilergym:cBench-v0/dijkstra> action -mem2reg
Action -mem2reg
Reward: 0.625806
Actions -mem2reg in 1.1ms with reward 0.6258064516129033.

...

compilergym:cBench-v0/dijkstra> stack
   Depth | Action   | Effect   | Done   |    Reward |   Cumulative Reward
---------+----------+----------+--------+-----------+---------------------
       8 | -reg2mem | True     | False  | -0.664516 |           -0.193548
       7 | -mem2reg | True     | False  |  0.625806 |            0.470968
       6 | -reg2mem | True     | False  | -0.664516 |           -0.154839
       5 | -mem2reg | True     | False  |  0.625806 |            0.509677
       4 | -reg2mem | True     | False  | -0.664516 |           -0.116129
       3 | -mem2reg | True     | False  |  0.625806 |            0.548387
       2 | -reg2mem | True     | False  | -0.703226 |           -0.077419
       1 | -mem2reg | True     | False  |  0.625806 |            0.625806
       0 | <init>   | False    | False  |  0        |            0
```

It could be that `remove_no_change()` could detect and remove these redundant subsequences.

Of course this is well beyond the scope of this PR so feel free to ignore me :)

ChrisCummins (commented):
Thanks for this @JD-ETH! This LGTM, with a couple of minor nitpicks. Aside from the CLA check, there's also some code formatters that will complain if you don't meet their pedantic criteria. You may want to install the [pre-commit hook](https://github.com/facebookresearch/CompilerGym/blob/development/.pre-commit-config.yaml#L4-L30) so that they will automatically format the code for you.

I have some thoughts on what constitutes an "effective action". I've left a rambly comment below. Nothing that needs addressing for this PR, but if you're interested in pursuing this further, I think a `minimize_action_sequence(env)`-type function would be a really useful addition to the LLVM API (it'll need to be compiler-specific as it will need to know about the internal IR state).

Cheers,
Chris

examples/random_walk.py
hughleat: Aye,

I'd like there to be `strict` mode for `simplify-stack` in the shell. If you end up building something like that, can you let me know?
If there is no reward set, it should pull the IRs and do the hashing like Chris says.  I'm not 100% sure that even that is enough - does LLVM keep anything sneaky anywhere? Still can't do anything about it if it does.
If there is a reward set, it should remove actions while that doing so increases or does not change the final reward. The current 'non strict' version just removes any passes that didn't have a positive reward last time. This isn't the same thing though, since a negative pass may be needed to set up a positive pass later. I was thinking a greedy approach would be sufficient, with multiple runs possible since one round of simplification changes the stack and may allow another run to have more effect (This definitely happens already with the non strict version).

[totally unimportant point] Chris, btw, 
`action -mem2reg -reg2mem -mem2reg -reg2mem -mem2reg -reg2mem -mem2reg -reg2mem`
Does the same thing as the 8 calls to action.

JD-ETH:
I tried to follow @ChrisCummins suggestion and updated the function accordingly. 
If you think this can go to the API in a similar form, I'm happy to update it and add some tests as long as you point me to the correct places. 
Screenshot 2021-02-18 at 22 36 48

For now I had some minitests to check the cancelling part is working. Would this qualify as the "strict" version?

examples/random_walk.py
```diff
@@ -44,20 +32,25 @@
...
 
     :param env: The environment to use.
     :param step_count: The number of steps to run. This value is an upper bound -
         fewer steps will be performed if any of the actions lead the
         environment to end the episode.
-    :return: The list of observed rewards.
     """
-    rewards = []
+
+    "Use hashes of the environment to record its internal state"
```
ChrisCummins: Nitpick:

```suggestion
    # Use hashes of the environment to record its internal state
```

(and for other comments in this file)

JD-ETH:
Seems like some tests are broken, unrelated to this PR?

examples/actor_critic.py
```diff
@@ -120,7 +120,7 @@ def __init__(self):
         self._env = gym.make("llvm-v0", reward_space=FLAGS.reward)
         try:
             self._env.require_dataset("cBench-v0")
-            self._env.reset(benchmark=FLAGS.benchmark)
+            self._env.reset(benchmark="benchmark://" + FLAGS.benchmark)
```
ChrisCummins: I think you had removed this in an earlier revision, perhaps you reverted that?

examples/random_walk.py
ChrisCummins: This is back :)

ChrisCummins (changes requested):
Looking good thanks! Just a couple of tiny nitpicks that I think you may have reverted by accident

Cheers,
Chris

examples/random_walk.py
ChrisCummins: Ignore this, this on an old diff! :)

JD-ETH:
Hey sorry I missed those comments before. Not really used to reverting commits :D

ChrisCummins (approved):
No worries, thanks for this! LGTM 👍 

Cheers,
Chris

PR #84, opened 2021-02-18
Run env.fork() on the service-side.

ChrisCummins:
Move the `fork()` implementation into a service-side API call, allowing a single service to share multiple connections to frontends. This new implementation is between a 44x and 77x faster, depending on the size of the benchmark being forked. `env.fork()` is now quicker than a single `env.step()` for the most part.

image

This also fixes a bug wherein calling `env.fork()` would cause rewards to diverge between environments.


tests/fuzzing/llvm_fork_env_fuzz_test.py
```diff
@@ -33,26 +34,29 @@ def test_fuzz(env: LlvmEnv):
             break
     else:
         # Fork the environment and check that the states are equivalent.
-        new_env = env.fork()
+        fkd = env.fork()
```
hughleat, line 37: I love the variable name :-)

tests/fuzzing/llvm_fork_env_fuzz_test.py
```diff
@@ -33,26 +34,29 @@ def test_fuzz(env: LlvmEnv):
...
                 action = env.action_space.sample()
                 observation_a, reward_a, done_a, _ = env.step(action)
-                observation_b, reward_b, done_b, _ = new_env.step(action)
+                observation_b, reward_b, done_b, _ = fkd.step(action)
 
                 print(env.state)  # For debugging in case of error.
-                np.testing.assert_array_almost_equal(observation_a, observation_b)
-                np.testing.assert_almost_equal(reward_a, reward_b)
                 assert done_a == done_b
+                np.testing.assert_array_almost_equal(observation_a, observation_b)
```
hughleat, line 50: Why `almost_equal`? None of the observations are noisy, yet, right?

hughleat (commented):
LGTM 👍

tests/fuzzing/llvm_fork_env_fuzz_test.py
ChrisCummins: Ah, good catch, force of habit :)

PR #86, opened 2021-02-19
added questions related to compiler as an environment

JD-ETH:
Added some FAQs related to using compilergym as RL env. 

BTW, what do you use to generate doc for local testing? I was manually running rst2html for a single file in this case and I doubt this is ideal. Never used RST before, thought it's rust at the beginning :D


ChrisCummins:
Thanks for this @JD-ETH! I think the wording could probably do with a bit of tightening up to make better fit the style of the rest of the documentation. Do you want to have a go at that?

To build the docs site locally I use:

```
make livedocs
```

you will need to run `make init` once to perform one-time setup. Run `make help` for more details.

Cheers,
Chris

JD-ETH:
Does this look more consistent now?

ChrisCummins:
Brilliant, thanks @JD-ETH ! Merging :)

Cheers,
Chris

PR #125, opened 2021-03-10
Tabular Q solution example

JD-ETH:
@ChrisCummins 
A working example with tabular Q-learning.

As this is my first real PR please be harsh about formatting, logging, commenting, styling, and point me to guidelines that I have missed. Thanks!


examples/brute_force.py
```diff
@@ -288,11 +306,15 @@ def run_brute_force(
...
 
     num_trials = sum(worker.num_trials for worker in workers)
+    env: CompilerEnv = make_env()
+    env.reset()
     print(
         f"completed {humanize.intcomma(num_trials)} of "
         f"{humanize.intcomma(expected_trial_count)} trials "
-        f"({num_trials / expected_trial_count:.3%})"
+        f"({num_trials / expected_trial_count:.3%}), best sequence",
+        ",".join([env.action_space.flags[i] for i in best_action_sequence]),
```
ChrisCummins:
```suggestion
        " ".join([env.action_space.flags[i] for i in best_action_sequence]),
```

Whitespace is a good separator for flags, as that is what you would use on the command line

examples/brute_force.py
```diff
@@ -288,11 +306,15 @@ def run_brute_force(
         worker.join()
 
     num_trials = sum(worker.num_trials for worker in workers)
+    env: CompilerEnv = make_env()
+    env.reset()
```
ChrisCummins:
```suggestion
```

Nitpick: no need to `reset()` here.

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+    FLAGS.reward = "IrInstructionCount"
+    FLAGS.benchmark = "cBench-v0/dijkstra"
+    FLAGS.env = "llvm-ic-v0"
+
+
+def main(argv):
+    setup_env()
+    # Train a Q table.
+    q_table = {}
+    env = get_env()
```
ChrisCummins: I would suggest using the `gym.make(...)` interface to set up the environment rather than setting commandline flags at runtime. If we use the environment `"llvm-autophase-ic-v0"` this already has the observation and reward spaces set up how you want. IMO this is easier to read, and also makes it clear to the user which flags can and cannot be overriden (now, only the `--benchmark` can be set):

```suggestion
    q_table: Dict[StateActionTuple, Reward] = {}
    benchmark = benchmark_from_flags()
    assert benchmark, "You must specify a benchmark using the --benchmark flag"
    env = gym.make("llvm-autophase-ic-v0", benchmark=benchmark)
```

I also added type hints to the `q_table` because I couldn't help myself 😃 If you go with this then the `setup_env()` and `get_env()` functions are not required, along with the `env_from_flags` import.

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+    try:
+        started = time()
+        train(q_table, env)
+        print(f"Time spent training {humanize.naturaldelta(time() - started)}")
```
ChrisCummins: Not a big deal, but there is a convenience class `compiler_gym.util.timer.Timer` to do this:

```suggestion
        with Timer("Constructed Q-table"):
            train(q_table, env)
```

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+        # Rollout based on the Max-Q policy.
+        ob = env.reset()
```
ChrisCummins: Nitpick: Prefer descriptive name `observation` over `ob` (and same for `action` instead of `a` below, and any others)

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+            q_table[hashed] = 0
+        max_q = max(q_table[hashed], max_q)
+    return max_q
+
+
+def train(q_table, env):
+    # Buffer an old version of q table to inspect training progress
+    prev_q = {}
+
+    for i in range(FLAGS.episodes):
```
ChrisCummins:
```suggestion
    for i in range(1, FLAGS.episodes + 1):
```

Super-nit ;-)

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+            target = reward + FLAGS.discount * get_max_q_value(
+                q_table, obs, current_length
+            )
+            # Update Q value at current state action pair.
+            q_table[hashed] = (
+                FLAGS.learning_rate * target
+                + (1 - FLAGS.learning_rate) * q_table[hashed]
+            )
+
+        if i % 50 == 0:
```
ChrisCummins: Not required, but it would be if there was a `FLAGS.log_every` flag which defaulted to 50

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+            print(f"Running episode {i}")
+
+            def compare_qs(q_old, q_new):
+                diff = [q_new[k] - v for k, v in q_old.items()]
+                return sum(diff) / len(diff) if diff else "NaN"
+
+            difference = compare_qs(prev_q, q_table)
+            print(
+                f"Newly added Q entries {len(q_table)-len(prev_q)}, averaged diff {difference}"
```
ChrisCummins: As well as logging the progression of the Q-table size/values, would it make sense to log some kind of measure of the quality of the trajectories evaluated? I had a little play around with buffering a batch of rewards and logging the average, maybe something like this would be nice:

```
episode=  50, avg_reward=0.6836, Q-table_entries= 1131, Q-table_diff=0.0015929
episode= 100, avg_reward=0.8026, Q-table_entries= 1945, Q-table_diff=0.0001957
episode= 150, avg_reward=0.8455, Q-table_entries= 2428, Q-table_diff=0.0002611
episode= 200, avg_reward=0.8591, Q-table_entries= 2803, Q-table_diff=0.0002467
...
episode=4950, avg_reward=0.9705, Q-table_entries=10365, Q-table_diff=0.0000241
episode=5000, avg_reward=0.9422, Q-table_entries=10380, Q-table_diff=0.0000213
```

done using:

```py
    prev_q = {}
    rewards = []
    for i in range(1, FLAGS.episodes + 1):
        observation = env.reset()
        for step in range(FLAGS.episode_length):
            # ...

        rewards.append(env.episode_reward)
        if i % FLAGS.log_every == 0:
            # ...
            avg_reward = sum(rewards) / len(rewards)
            rewards = []
            print(
                f"episode={i:4d}, avg_reward={avg_reward:.4f}, "
                f"Q-table_entries={len(q_table):5d}, Q-table_diff={difference:.7f}"
            )
```

I'm not sure if reward average makes sense, and either way it isn't a big deal at all

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+            "Resulting sequence: ", ",".join(action_seq), f"total reward {sum(rewards)}"
+        )
+        # Rollout the best episode from brute force solution.
+        best = ["-gvn-hoist", "-newgvn", "-instcombine", "-mem2reg", "-simplifycfg"]
+        rewards = []
+        env.reset()
+        for a in best:
+            _, r, _, _ = env.step(env.action_space.flags.index(a))
+            rewards.append(r)
+        print(f"Best achievable result {sum(rewards)}")
```
ChrisCummins:
```suggestion
```

I don't think we should include a hardcoded policy. I understand why it's useful to compare against a global optimum, but this is too fragile, and will break if a user tries a different benchmark/reward etc. I think we leave it as an exercise to the reader to figure out how good the solution is :)

examples/tabular_q.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def hash_state_action(autophase_feature, action, step):
+    return tuple(
+        (
+            tuple(autophase_feature[FLAGS.features_indices]),
+            step,
+            FLAGS.actions.index(action),
+        )
+    )
```
ChrisCummins: In general I would prefer there to be a bit more documentation for the functions in this file. In particular, this one, since, even if the reader is a compiler/rl expert, this function contains both implementation details (the need to create a hashable tuple), and design decisions (the choice of autophase feature values), etc.

I would suggest something like this (though note it is an incomplete scripbble :-) ):

```suggestion
def make_q_table_key(autophase_feature, action, step):
    """Create a hashable Q-table key.

    For tabular learning we will be constructing a Q-table which maps a
    (state, action) pair to an expected reward. The purpose of this function is
    to convert the (state, action) properties into a hashable tuple that can be
    used as a key for a Q-table dictionary.

    The (state, action) tuple returned by this function consists of ... (TODO:
    explain the choice of autophase features, the purpose of step count, etc).
    """
    return *autophase_feature[FLAGS.features_indices], step, FLAGS.actions.index(action)
```

I would point to [finish_episode()](https://github.com/facebookresearch/CompilerGym/blob/development/examples/actor_critic.py#L240-L307) as a guide for figuring out the level of detail required when documenting non-obvious code

examples/brute_force.py
```diff
@@ -43,7 +43,23 @@
...
+        "-instcombine",
+        "-jump-threading",
+        "-loop-extract",
+        "-loop-reduce",
+        "-loop-versioning",
+        "-newgvn",
+        "-mem2reg",
+        "-sroa",
+        "-structurizecfg",
+    ],
```
ChrisCummins: I would prefer to leave the default value as an empty list. This script is not specific to LLVM, and atm an empty list will auto-select all of the actions available in the environment

examples/brute_force.py
```diff
@@ -262,15 +278,17 @@ def run_brute_force(
...
+                    f"Best reward found: {best_reward}.",
                     file=sys.stderr,
                     flush=True,
                     end="",
                 )
                 for actions, rewards in chunk:
                     print(*actions, *rewards, sep=",", file=f, flush=True)
                     if rewards and rewards[-1] is not None:
-                        best_reward = max(best_reward, rewards[-1])
+                        if sum(rewards) > best_reward:
```
ChrisCummins, line 273: Nice :-) Thanks for the fix!

ChrisCummins (changes requested):
This really nice @JD-ETH, thanks a lot! Having a value-iteration baseline is a very welcome addition.

In general the implementation is solid, but *since you asked*, I have gone through and left a very pedantic review 😃 Don't worry, this level of scrutiny is not normal! Most things are small nitpicks or bikeshedding. The most important change I'd love to see is to bring the level of commenting up to the point of this being a reasonably-standalone explanation of the technique for a general non-expert audience.

Leave a comment here once you want me to take another look.

Cheers,
Chris

ChrisCummins:
Oo one thing I forgot - please add a test 😊

Given that this isn't core API, a simple smoke test using flags for quick run like `--episodes=5` would be sufficient. Take a look at the actor critic smoke test as an example: https://github.com/facebookresearch/CompilerGym/blob/development/examples/actor_critic_test.py

You will then need to add bazel definitions so that the test is run on `make test`: https://github.com/facebookresearch/CompilerGym/blob/development/examples/BUILD#L8-L25

EDIT: Also worth noting that this test will fail until this PR is rebased on #126.

examples/tabular_q.py
JD-ETH: Typically in RL one would *evaluate* the policy by rolling out in the environment and observe its performance. I'm adding a rollout function and reporting the resulting rewards of the trajectory. It's a good point though.

JD-ETH:
@ChrisCummins  I'm ready for the next round.

examples/tabular_q.py
ChrisCummins: Sounds good, thanks!

examples/tabular_q.py
```diff
@@ -0,0 +1,206 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""Simple compiler gym tabular q learning example.
+Usage python tabular_q.py
```
ChrisCummins:
```suggestion
Usage python tabular_q.py --benchmark=<benchmark>
```

Tiny nit :)

ChrisCummins (approved):
Hi @JD-ETH, thanks a lot for making those changes, I really appreciate the extra docs and the test!

This LGTM and feel free to merge it when you're ready. You may want to squash the history before merging into smaller atomic commits, perhaps one for the brute_force fix and the other for the tabulary_q, but that's not essential so I'll leave it up to your discretion :) Thanks for adding this new baseline!

Cheers,
Chris

examples/tabular_q.py
```diff
@@ -0,0 +1,206 @@
...
+    return FLAGS.actions[random.choice(max_indices)]
+
+
+def get_max_q_value(q_table, ob, step):
+    max_q = 0
+    for act in FLAGS.actions:
+        hashed = make_q_table_key(ob, act, step)
+        if hashed not in q_table:
+            q_table[hashed] = 0
+        max_q = max(q_table[hashed], max_q)
```
benoitsteiner: You could use the more pythonic, more compact and faster max_q = max(q_table.get(hashed, 0), max_q) instead of inserting 0 in the q_table. You just need to make sure to use it consistently throughout the code: since hashed will not be inserted in the hashtable, subsequent calls to q_table[hashed] may throw unless you replace them with calls to  q_table.get(hashed, 0)

JD-ETH: Thanks for the input!

JD-ETH:
@ChrisCummins Now i see CI is failing from this PR, sorry for merging it before the CI tests finished. I am not sure I understand what failed in the test though, would you give some pointers?

ChrisCummins:
> @ChrisCummins Now i see CI is failing from this PR, sorry for merging it before the CI tests finished. I am not sure I understand what failed in the test though, would you give some pointers?

Thanks for checking. It's a false positive, nothing to worry about :). See #144.

PR #126, opened 2021-03-11
[llvm] Fix initial observation value on env.reset().

ChrisCummins:
Fixes a regression introduced in 6cccf05e9004a2fff5bb15c95afec1de8b897062 that meant that env.reset() would not return an observation value for `LlvmEnv` instances.


JD-ETH:
Looks fine to me, but do you know what's failing the tests?

JD-ETH (approved):
LGTM when the tests pass.

ChrisCummins:
Thanks @JD-ETH. I bumped `cBench-v0` to `cBench-v1` but didn't update this test. Will do that now

Issue #46, opened 2021-01-28
[llvm] -gvn-sink action produces nondeterministic states


ChrisCummins:
## 🐛 Bug

The `-gvn-sink` action has been found to produce different states when run multiple times on the same input.

## To Reproduce

1. Create an episode that runs `-gvn-sink`:
```sh
cat <<EOF > states.csv
benchmark,reward,walltime,commandline
benchmark://cBench-v0/blowfish,0.35070822,,opt -gvn-sink input.bc -o output.bc
EOF
```
2. Run the episode:
```sh
python -m compiler_gym.bin.validate --env=llvm-ic-v0 < states.csv
```
3. Check the output reported by the validation script. Sometimes it succeeds:

```
$ python -m compiler_gym.bin.validate --env=llvm-ic-v0 < states.csv
✅  benchmark://cBench-v0/blowfish  0.3507
```

Sometimes it fails:
```
$ python -m compiler_gym.bin.validate --env=llvm-ic-v0 < states.csv
❌  benchmark://cBench-v0/blowfish  Expected reward 0.3507 but received reward 0.3547
```

## Expected behavior

The `-gvn-action` should reliably produce the same state.

## Environment

Please fill in this checklist:

- CompilerGym: 0.1.2
- How you installed CompilerGym (conda, pip, source): pip
- OS: Linux
- Python version: 3.8

PR #127, opened 2021-03-11
[llvm-ic-v0/cBench-v1] Leaderboard Submission: Greedy search

ChrisCummins:
This adds a greedy search policy that at each step will choose to select an action
randomly with probability ε.

At each step the agent selects either a greedy policy or an exploration policy.
If greedy, every action is evaluated and the action with the greatest reward
is selected. If exploration, an action is selected randomly. The episode
terminates when the maximum reward attainable by any action is <= 0.

## What is in this PR?

* An implementation of an e-greedy agent, using the [eval_policy()](https://github.com/facebookresearch/CompilerGym/blob/development/leaderboard/llvm_codesize/eval_policy.py) helper to run it on the codesize reduction task.
* A write-up of the approach, based on the [leaderboard submission template](https://github.com/facebookresearch/CompilerGym/blob/development/leaderboard/SUBMISSION_TEMPLATE.md).
* A leaderboard entry of results achieved using a pure greedy approach, verified using:

```
$ python -m compiler_gym.bin.validate --env=llvm-ic-v0 results_e0.csv
✅  cBench-v0/crc32                           1.0000
✅  cBench-v0/crc32                           1.0000
...
✅  cBench-v0/stringsearch2                   0.9742
✅  cBench-v0/ghostscript                     1.0016
----------------------------------------------------
Number of validated results: 230 of 230
Mean walltime per benchmark: 145.705s (std: 380.377s)
Geometric mean IrInstructionCountOz: 1.042 (std: 0.116)
```

and result achieved using epsilon=0.10 exploratory approach:

```
$ python -m compiler_gym.bin.validate --env=llvm-ic-v0 results_e10.csv
✅  cBench-v0/adpcm                           1.0000
...
✅  cBench-v0/tiffmedian                      1.0021
✅  cBench-v0/tiffmedian                      1.0090
----------------------------------------------------
Number of validated results: 230 of 230
Mean walltime per benchmark: 152.579s (std: 437.295s)
Geometric mean IrInstructionCountOz: 1.041 (std: 0.117)
```


leaderboard/llvm_codesize/e_greedy/e_greedy.py
```diff
@@ -0,0 +1,115 @@
...
+    # using action index as a tie-breaker. Do this by creating n forks of the
+    # environment, one for every action, and evaluting the actions in parallel
+    # threads. Note that calls to fork() occur in the main thread for thread
+    # safety in case of environment restart.
+    futures = (
+        executor.submit(eval_action, env.fork(), action)
+        for action in range(env.action_space.n)
+    )
+    results = (future.result() for future in as_completed(futures))
+    results = sorted(
```
JD-ETH: Could have been faster if just scanning through the list and keep the best instead of sorting.

leaderboard/llvm_codesize/e_greedy/e_greedy.py
```diff
@@ -0,0 +1,115 @@
...
+        finally:
+            fkd.close()
+        return ActionReward(action=action, reward=reward)
+
+    # Create an ordered list of actions, best first, ordered by descding reward,
+    # using action index as a tie-breaker. Do this by creating n forks of the
+    # environment, one for every action, and evaluting the actions in parallel
+    # threads. Note that calls to fork() occur in the main thread for thread
+    # safety in case of environment restart.
+    futures = (
```
JD-ETH, line 51: This is fancy and looks cool!

However I wonder if it'd be faster to implement a revert_step() for the environment to return to previous state? Maybe compared to forking another environment could be faster? 

But obviously forking is a more general option.

leaderboard/llvm_codesize/e_greedy/e_greedy.py
```diff
@@ -0,0 +1,115 @@
...
+                    logging.debug(
+                        "Step %d, greedy action %s, reward %.4f, cumulative %.4f",
+                        step_count,
+                        env.action_space.flags[best.action],
+                        reward,
+                        env.episode_reward,
+                    )
+                    if env.reward_space.deterministic and reward != best.reward:
+                        logging.warning(
+                            "Action %s produced different reward on replay, %.4f != %.4f",
```
JD-ETH, line 104: How is this possible?

leaderboard/llvm_codesize/e_greedy/README.md
```diff
@@ -0,0 +1,49 @@
+# Epsilon-Greedy Search
+
+**tldr;**
+A greedy search policy that at each step will choose to select an action
```
JD-ETH: This doesn't seem to mention your `greedy` policy, especially that you are only rolling out the trajectory by **one** step and picking the best.

JD-ETH (approved):
This is a great example for anyone else to follow. Left minor comments and it looks good otherwise.

leaderboard/llvm_codesize/e_greedy/e_greedy.py
ChrisCummins: This is possible if one of the underlying actions is nondeterministic. This has been found in the past, e.g.
https://github.com/facebookresearch/CompilerGym/issues/46

leaderboard/llvm_codesize/e_greedy/e_greedy.py
ChrisCummins: Good point! Done :)

leaderboard/llvm_codesize/e_greedy/e_greedy.py
ChrisCummins: Yeah a `revert_step()` would be super useful, but the problem is that the compiler passes don't have an "undo" function, so the only way to do this would be to call `env.fork()` before every step, building up a stack of environments, one for each step, and to pop from the top on `revert_step()`. This is totally do-able btw

leaderboard/llvm_codesize/e_greedy/README.md
ChrisCummins: Thanks, good spot! Fixed

benoitsteiner (approved):
Nice work

leaderboard/llvm_codesize/e_greedy/e_greedy.py
```diff
@@ -0,0 +1,116 @@
...
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from typing import NamedTuple
+
+from absl import flags
+
+from compiler_gym.envs import CompilerEnv, LlvmEnv
+
+# Import the ../eval_policy.py helper.
+sys.path.insert(0, os.path.dirname(os.path.realpath(__file__)) + "/..")
+from eval_policy import eval_policy  # noqa pylint: disable=wrong-import-position
```
JD-ETH, line 19: Is there a better way to include this? Probably people would want to copy this as an example. Should I open an issue?

ChrisCummins: Yes, please feel free to open an issue. We could move `eval_policy()` into the compiler_gym package so that so people could use something like:

```py
from compiler_gym.leaderboard.llvm_codesize import eval_policy
```

ChrisCummins: Filed #158 to follow up on this.

PR #128, opened 2021-03-11
#111 validation patch set.

ChrisCummins:
Split out patches from #111:

* Removes several actions from the LLVM action space that change the program semantics (e.g. a pass that extracts loops into their own functions, or a pass that inserts runtime bounds checks).
* Fixes a couple of small logging bugs.
* Some small fixes to improve error handling.


ChrisCummins:
Heads up @JD-ETH that this updates the pass list used by actor critic. These changes will need to reflected in the new Q-tabular implementation.

Cheers,
Chris

JD-ETH (approved):
Hey Christ, thanks for asking me to review. 
I assume the purpose of the request is to make aware that there are changes in the action space definition of CompilerGym during your validation investigation, as I'm certainly not qualified to understand the codebase. 
Given that assumption and seeing the many tests you are adding I am approving :D

tests/llvm/fork_env_test.py
```diff
@@ -90,9 +90,12 @@ def test_fork_before_reset(env: LlvmEnv):
...
     fkd = env.fork()
-    assert env.in_episode
-    assert fkd.in_episode
-    assert env.benchmark == fkd.benchmark
+    try:
+        assert env.in_episode
+        assert fkd.in_episode
+        assert env.benchmark == fkd.benchmark
+    finally:
+        fkd.close()
```
JD-ETH, line 98: Just wondering, is the assertion fails does the error gets propagated past `finally` without catching exception?

ChrisCummins: Yes, `finally` does not trap the error, it just ensures that `fkd.close()` is run even if there is an error. Demo:

```py
>>> try:
...     assert False
... finally:
...     print("Hi, finally!")
...
Hi, finally!
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
AssertionError
```

PR #130, opened 2021-03-15
Add an example service implementation in Python

ChrisCummins:
This extends the `//examples/example_compiler_gym_service` directory with a new service, implemented in Python, with exact feature parity of the C++ implementation.


examples/example_compiler_gym_service/__init__.py
```diff
@@ -8,46 +8,58 @@
 from compiler_gym.util.runfiles_path import runfiles_path
 
 
-class CodesizeReward(Reward):
-    """An example reward that uses changes in the "codesize" observation value
+class RuntimeReward(Reward):
```
bwasti, line 11: I wonder if this could be even more thin -- seems like the pertinent info could be passed into the Reward() constructor (deterministic, platform dependent, etc) and everything is just a scalar of some kind by default

examples/example_compiler_gym_service/service_py/example_service.py
```diff
@@ -0,0 +1,225 @@
...
+                observation.string_value = "Hello, world!"
+            elif observation_space == 1:  # features
+                observation.int64_list.value[:] = [0, 0, 0]
+            elif observation_space == 1:  # runtime
+                observation.scalar_double = 0
+
+        return reply
+
+
+class ExampleCompilerGymService(proto.CompilerGymServiceServicer):
```
bwasti, line 122: this could probably be hidden away and take in a class that implements `CompilationSession` (with the addition of class methods to get `ACTION_SPACE` and `OBSERVATION_SPACES`)

bwasti (approved):
Light comments on design :)

thanks for the quick demo!

examples/example_compiler_gym_service/__init__.py
ChrisCummins: Good idea. This is kind of how the LLVM backend does it ([e.g.](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/llvm_env.py#L84-L91)), but I haven't tried to generalize it yet. Filing #131 to follow up

examples/example_compiler_gym_service/service_py/example_service.py
ChrisCummins: I agree, this would be really nice. I think once we start adding support for multiple compilers we should refactor away the boilerplate for session management / handling RPC requests

ChrisCummins:
Thanks for the feedback @bwasti!

PR #133, opened 2021-03-17
Add trajectory minimization utilities

ChrisCummins:
This adds a set of utility functions for minimizing trajectories.

A trajectory is the sequence of actions applied to an environment. The goal of
trajectory minimization is to identify the shortest subregion of a trajectory
such that some hypothesis still holds. A a hypothesis is a boolean test on an
environment.

## Example usage

Finding the first action that causes validation to fail:

```py
bisect_trajectory(env, hypothesis=lambda env: not env.validate().okay())
```

Using random pruning to attempt to reduce the number of actions needed to reach some target reward:

```py
random_minimization(env, hypothesis=lambda env: env.episode_reward >= target_reward)
```

### DIFF
--- a/compiler_gym/util/minimize_trajectory.py
+++ b/compiler_gym/util/minimize_trajectory.py
@@ -1,0 +1,329 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""This module defines utilities for minimizing trajectories.
+
+A trajectory is the sequence of actions applied to an environment. The goal of
+trajectory minimization is to identify the shortest subregion of a trajectory
+such that some hypothesis still holds. A a hypothesis is a boolean test on an
+environment, for example, a hypothesis could be that :code:`env.validate()`
+returns an error.
+"""
+import logging
+import random
+from math import ceil, log
+from typing import Callable, Iterable
+
+from compiler_gym.envs import CompilerEnv
+from compiler_gym.util.truncate import truncate
+
+
+class MinimizationError(OSError):
+    """Error raised if trajectory minimization fails."""
+
+
+# A hypothesis is a callback that accepts as input an enivornment in a given
+# state returns true if a particular hypothesis holds, else false.
+Hypothesis = Callable[[CompilerEnv], bool]
+
+
+def environment_validation_fails(env: CompilerEnv) -> bool:
+    """A hypothesis that holds true if environment validation fails."""
+    validation_result = env.validate()
+    logging.debug(truncate(str(validation_result), max_lines=1, max_line_len=120))
+    return not validation_result.okay()
+
+
+def _apply_and_test(env, actions, hypothesis, flakiness) -> bool:
+    """Run specific actions on environment and return whether hypothesis holds."""
+    env.reset(benchmark=env.benchmark)
+    for _ in range(flakiness):
+        logging.debug("Applying %d actions ...", len(actions))
+        _, _, done, info = env.step(actions)
+        if done:
+            raise MinimizationError(
+                f"Failed to replay actions: {info.get('error_details', '')}"
+            )
+        logging.debug("Applied %d actions", len(actions))
+        if hypothesis(env):
+            return True
+    return False
+
+
+def bisect_trajectory(
+    env: CompilerEnv,
+    hypothesis: Hypothesis = environment_validation_fails,
+    reverse: bool = False,
+    flakiness: int = 1,
+) -> Iterable[CompilerEnv]:
+    """Run a binary search to remove the suffix or prefix of a trjectory.
+
+    Requires worst-case O(log n) evaluation rounds, where n is the length of the
+    trajectory.
+
+    :param env: An environment whose action trajectory should be minimized.
+    :param hypothesis: The hypothesis that is used to determine if a trajectory
+        is valid. A callback that accepts as argument the :code:`env:`
+        instance and returns true if the hypothesis holds, else false. The
+        hypothesis must hold on the initial trajectory.
+    :param reverse: If :code:`True`, minimize the trajectory from the front
+        (i.e. the prefix). Else, minimization occurs form the back (i.e. the
+        suffix).
+    :param flakiness: The maximum number of times the hypothesis is repeated
+        to check if it holds. If the hypothesis returns :code:`True` within this
+        many iterations, it is said to hold. It needs to only return
+        :code:`True` once.
+    :returns: A generator that yields the input environment every time the
+        trajectory is successfully reduced.
+    :raises MinimizationError: If the environment action replay fails, or if
+        the hypothesis does not hold on the initial trajectory.
+    """
+
+    def apply_and_test(actions):
+        return _apply_and_test(env, actions, hypothesis, flakiness)
+
+    all_actions = env.actions.copy()
+    # No actions to minimize.
+    if not all_actions:
+        return env
+
+    logging.info(
+        "%sisecting sequence of %d actions",
+        "Reverse b" if reverse else "B",
+        len(all_actions),
+    )
+    if not apply_and_test(all_actions):
+        raise MinimizationError(
+            "Hypothesis failed on the initial state! The hypothesis must hold for the first state."
+        )
+
+    left = 0
+    right = len(all_actions) - 1
+    step = 0
+    while right >= left:
+        step += 1
+        remaining_steps = int(log(max(right - left, 1), 2))
+        mid = left + ((right - left) // 2)
+        logging.debug(
+            "Bisect step=%d, left=%d, right=%d, mid=%d", step, left, right, mid
+        )
+
+        actions = all_actions[mid:] if reverse else all_actions[:mid]
+        if apply_and_test(actions):
+            logging.info(
+                "🟢 Hypothesis holds at num_actions=%d, remaining bisect steps=%d",
+                mid,
+                remaining_steps,
+            )
+            yield env
+            if reverse:
+                left = mid + 1
+            else:
+                right = mid - 1
+        else:
+            logging.info(
+                "🔴 Hypothesis does not hold at num_actions=%d, remaining bisect steps=%d",
+                mid,
+                remaining_steps,
+            )
+            if reverse:
+                right = mid - 1
+            else:
+                left = mid + 1
+
+    mid = max(left, right) - 1 if reverse else min(left, right) + 1
+    if (reverse and mid < 0) or (not reverse and mid >= len(all_actions)):
+        actions = all_actions
+        logging.info("Failed to reduce trajectory length using bisection")
+    else:
+        actions = all_actions[mid:] if reverse else all_actions[:mid]
+        logging.info(
+            "Determined that action %d of %d is the first at which the hypothesis holds: %s",
+            mid,
+            len(all_actions),
+            env.action_space.flags[all_actions[mid]],
+        )
+
+    if not apply_and_test(actions):
+        raise MinimizationError("Post-bisect sanity check failed!")
+
+    yield env
+
+
+def random_minimization(
+    env: CompilerEnv,
+    hypothesis: Hypothesis = environment_validation_fails,
+    num_steps_ratio_multiplier: float = 5,
+    init_discard_ratio: float = 0.75,
+    discard_ratio_decay: float = 0.75,
+    min_trajectory_len: int = 5,
+    flakiness: int = 1,
+) -> Iterable[CompilerEnv]:
+    """Run an iterative process of randomly removing actions to minimize a
+    trajectory.
+
+    For each round of minimization, a number of actions are discarded randomly
+    and the hypothesis is tested. If the hypothesis still holds with those
+    actions removed, the minimization proceeds. Else the actions are re-inserted
+    into the trajectory and a new set of actions are removed. After a failure
+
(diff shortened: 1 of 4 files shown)


compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+
+# A hypothesis is a callback that accepts as input an enivornment in a given
+# state returns true if a particular hypothesis holds, else false.
+Hypothesis = Callable[[CompilerEnv], bool]
+
+
+def environment_validation_fails(env: CompilerEnv) -> bool:
+    """A hypothesis that holds true if environment validation fails."""
+    validation_result = env.validate()
+    logging.debug(str(validation_result).split("\n")[0][:120])
```
hughleat: Why is there 120 here?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    return not validation_result.okay()
+
+
+def _apply_and_test(env, actions, hypothesis, flakiness) -> bool:
+    """Run specifiec actions on environment and return whether hypothesis holds."""
```
hughleat: spelling

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    env.reset(benchmark=env.benchmark)
```
hughleat, line 40: Currious why this isn't just `env.reset()`

compiler_gym/util/minimize_trajectory.py
hughleat: Also, what is `flakiness`?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    for _ in range(flakiness):
+        logging.debug("Applying %d actions ...", len(actions))
+        _, _, done, info = env.step(actions)
+        if done:
+            raise MinimizationError(
+                f"Failed to replay actions: {info.get('error_details', '')}"
+            )
+        logging.debug("Applied %d actions", len(actions))
+        if hypothesis(env):
+            return True
```
hughleat, line 50: Interesting. Not sure I understand. Why is not returning if `hypothesis` is false once? Currently it is returning if `hypothesis` is true once.

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+
+    :param env: An environment whose action trajectory should be minimized.
+    :param hypothesis: The hypothesis that is used to determine if a trajectory
+        is valid. A callback that accepts as argument the :code:`env:`
+        instance and returns true if the hypothesis holds, else false. The
+        hypothesis must hold on the initial trajectory.
+    :param reverse: If :code:`True`, minimize the trajectory from the front
+        (i.e. the prefix). Else, minimization occurs form the back (i.e. the
+        suffix).
+    :param flakiness: The maximum number of times the hypothesis is repeated
```
hughleat, line 73: If it holds every time, or at least once?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+            raise MinimizationError(
+                f"Failed to replay actions: {info.get('error_details', '')}"
+            )
+        logging.debug("Applied %d actions", len(actions))
+        if hypothesis(env):
+            return True
+    return False
+
+
+def bisect_trajectory(
```
hughleat, line 54: I'm curious as to how often this technique helps. 
What about having a starting point in a sequence, trying to build the longest section you can snip from there by binary search. That way, you can start at the beginning, find the longest sequence you can remove, then include the next action and repeat from there. If everything but the prefix and suffix are needed, then it has the same effect as yours (but more costly). But if there are also big holes in the middle - which is presumably as likely as having them at the ends - then it will prune them too.

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+) -> Iterable[CompilerEnv]:
+    """Run an iterative process of randomly removing actions to minimize a
+    trajectory.
+
+    For each round of minimization, a number of actions are discarded randomly
+    and the hypothesis is tested. If the hypothesis still holds with those
+    actions removed, the minimization proceeds. Else the actions are re-inserted
+    into the trajectory and a new set of actions are removed. After a failure
+
+    Performs up to O(num_steps_ratio_multiplier * log n) evaluation rounds,
```
hughleat, line 171: At first blush, that seems unnecessarily complicated. Why not just perform n steps and accept n as a param?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    discard_ratio_decay: float = 0.75,
+    flakiness: int = 1,
+) -> Iterable[CompilerEnv]:
+    """Run an iterative process of randomly removing actions to minimize a
+    trajectory.
+
+    For each round of minimization, a number of actions are discarded randomly
+    and the hypothesis is tested. If the hypothesis still holds with those
+    actions removed, the minimization proceeds. Else the actions are re-inserted
+    into the trajectory and a new set of actions are removed. After a failure
```
hughleat, line 169: "After a failure..."?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    Performs up to O(num_steps_ratio_multiplier * log n) evaluation rounds,
+    where n is the length of the trajectory.
+
+    :param env: An environment whose action trajectory should be minimized.
+    :param hypothesis: The hypothesis that is used to determine if a trajectory
+        is valid. A callback that accepts as argument the :code:`env:`
+        instance and returns true if the hypothesis holds, else false. The
+        hypothesis must hold on the initial trajectory.
+    :param num_steps_ratio_multiplier: A multiplier for the number of rounds of
+        minimization to perform, using log(n) the length of the trajectory as
```
hughleat, line 180: log(length of the trajectory)?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+        the factor.
+    :param init_discard_ratio: The number of actions that will be randomly
```
hughleat, line 182: Maybe describe the exponential decay thing in the intro?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+        discarded, as a multiplier of the length of the trajectory.
+    :param discard_ratio_decay: The ratio of decay for the discard ratio on
+        failure.
+    :param flakiness: The maximum number of times the hypothesis is repeated
+        to check if it holds.
+    :returns: A generator that yields the input environment every time the
+        trajectory is successfully reduced.
+    :raises MinimizationError: If the minimization fails, or if the hypothesis
```
hughleat: BTW, how will minimisation fail? Do you mean it just doesn't yield anything? Isn't that just returning an empty generator?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    if not apply_and_test(actions):
+        raise ValueError(
+            "Hypothesis failed on the initial state! The hypothesis must hold for the first state."
+        )
+
+    max_num_steps = int(log(len(actions), 2) * num_steps_ratio_multiplier)
+
+    num_steps = 0
+    discard_ratio = init_discard_ratio
+    while len(actions) > 5 and num_steps < max_num_steps:
```
hughleat: Why `len(actions) > 5`?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+        num_steps += 1
+        num_to_remove = int(ceil(len(actions) * discard_ratio))
+        candidate_actions = actions.copy()
+        for _ in range(num_to_remove):
+            candidate_actions.remove(random.choice(candidate_actions))
```
hughleat: Not 100% sure how the underlying code works here. Is this removing by index or value?
If removing by value, then you might be removing all `reg2mem` for example. If there are repeated actions then that might not be what is expected.

compiler_gym/util/minimize_trajectory.py
hughleat: Actually, it is not easy to understand exactly what this code does from the doc. E.g. the `init_discard_ratio` is reset on success. The process that will be undertaken is not very clear, which makes it had to know what parameter values might be sensible.

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+            actions = candidate_actions
+            discard_ratio = init_discard_ratio
+            yield env
+        else:
+            logging.info(
+                "🔴 Hypothesis does not hold with %s of %s actions randomly removed, rolling back",
+                num_to_remove,
+                len(actions),
+            )
+            discard_ratio *= discard_ratio_decay
```
hughleat, line 234: I feel like I could just be unlucky. Can't I have some number of attempts at a given `discard_ratio`?

compiler_gym/util/minimize_trajectory.py
```diff
@@ -0,0 +1,317 @@
...
+    action_has_been_pruned = True
+    # Outer loop. Repeat iterative reduction until no change is made.
+    while action_has_been_pruned and len(all_actions) > 1:
+        pass_num += 1
+        action_has_been_pruned = False
+        action_mask = [True] * len(all_actions)
+        logging.info("Minimization pass on sequence of %d actions", len(all_actions))
+
+        # Inner loop. Go through every action and see if it can be removed.
+        for i in range(len(action_mask)):
```
hughleat, line 297: What's the mask for? Why not just delete the actions like in previous algs?

hughleat (approved):
LGTM. A few questions here or there.

compiler_gym/util/minimize_trajectory.py
ChrisCummins: That's just to stop crowding the screen if the error message is huge. Replaced it with a more clear `truncate()` call.

compiler_gym/util/minimize_trajectory.py
ChrisCummins: If a user doesn't set a specific benchmark, then calling `reset()` allows the environment to select a new benchmark randomly, e.g.:

```py
>>> env = gym.make("llvm-v0")
>>> env.reset()
>>> env.benchmark
'benchmark://cBench-v1/jpeg-c'
>>> env.reset()
>>> env.benchmark
'benchmark://cBench-v0/tiff2bw'
```

compiler_gym/util/minimize_trajectory.py
ChrisCummins: Yeah, that isn't clear :) I'm using "flakiness" in the same way I would if describing some test that is flaky - we allow it fail a couple of times, but it only needs to pass once. Will expand the comments

compiler_gym/util/minimize_trajectory.py
ChrisCummins: Good catch, fixed the comment

compiler_gym/util/minimize_trajectory.py
ChrisCummins: Just some magic constant for what "feels" like a good number to bottom out on. If you have less than 5 actions left you could just iterate through each of them rather than trying it by hand. I've replaced it with a `min_trajectory_len` param to be more clear

compiler_gym/util/minimize_trajectory.py
ChrisCummins: Ah, great catch! `remove()` is removing by value. So this means if there are duplicates, only the first instance is removed. I don't want to do this, so I'll replace it with an index-based approach.

compiler_gym/util/minimize_trajectory.py
ChrisCummins: The reason for only using a single attempt at each discard ratio is because I found that being super aggressive with the ratio (75%) and backing it off quickly was more effective than being more conservative and trying for more attempts. I might be dead wrong about here, it was just a hunch I picked up while playing around with a few different buggy trajectories that I was minimizing :)

compiler_gym/util/minimize_trajectory.py
ChrisCummins: IMO the mask is easier to read for the case where we're flicking actions on and off, but no particular reason :)

ChrisCummins:
Thanks a lot for the review @hughleat!

compiler_gym/util/minimize_trajectory.py
hughleat: That feels a bit odd. Does it always select a new benchmark?

hughleat:
I'd like to propose a different search. This first tries to delete all actions. If that doesn't work it tries to delete the first half, then second half. And so on recursively. Max cost is nlgn, but much smaller if there are large regions of dead actions.
```python
def recursive_binary_partition(actions, mask=ones, lo=0, hi=len(actions)):
  if lo == hi:
    return
  mask[lo:hi] = 0
  if hypothesis(masked actions):
    return
  mask[lo:hi] = 1
  md = (lo+hi)/2
  recursive_binary_partition(actions, mask, lo, md)
  recursive_binary_partition(actions, mask, md, hi)
```

PR #136, opened 2021-03-17
Add default rewards for non-LLVM backends

bwasti:
Removes some hard coding and provides the option of default rewards when they are unspecified in the CompilerEnv

### DIFF
--- a/compiler_gym/envs/compiler_env.py
+++ b/compiler_gym/envs/compiler_env.py
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 """This module defines the OpenAI gym interface for compilers."""
 import logging
+import numbers
 import os
 import sys
 import warnings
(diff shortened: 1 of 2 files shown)


compiler_gym/envs/compiler_env.py
```diff
@@ -294,7 +325,7 @@ def commandline(self) -> str:
 
         :return: A string commandline invocation.
         """
-        raise NotImplementedError("abstract method")
+        return "not implemented"
```
ChrisCummins: Why? :)

compiler_gym/random_search.py
```diff
@@ -164,7 +162,7 @@ def random_search(
 
     print(
         f"Started {len(workers)} worker threads for "
-        f"{benchmark_name} ({humanize.intcomma(metadata['num_instructions'])} instructions) "
+        f"{benchmark_name} ({humanize.intcomma(metadata['reward'])}) "
         f"using reward {reward_space_name}."
```
ChrisCummins: Nitpick: this prints out the rewards space name twice:

```
$ bazel run -c opt //compiler_gym/bin:random_search -- --env=llvm-ic-v0 --benchmark=cBench-v1/dijkstra
Started 24 worker threads for benchmark://cBench-v1/dijkstra (IrInstructionCountOz) using reward IrInstructionCountOz.
```

perhaps just:

```suggestion
        f"{benchmark_name} using reward {reward_space_name}."
```

compiler_gym/envs/compiler_env.py
```diff
@@ -209,6 +234,12 @@ def __init__(
             logger=self.logger,
         )
 
+        # If no reward space is specified, generate some from the observation space
+        rewards = rewards or [
+            DefaultRewardFromObservation(obs.name)
+            for obs in self.service.observation_spaces
```
ChrisCummins, line 241: We could get a bit fancy here and filter only the observation spaces that are scalar ints/floats, since `DefaultRewardFromObservation` would not work for a string/bytes etc observation type. Not required for this PR, though feel free to add a TODO note

ChrisCummins (approved):
Thanks for this @bwasti! LGTM, with a couple of minor nits left inline.

I like the addition of default rewards. Ever since https://github.com/facebookresearch/CompilerGym/pull/69, we broke the promise that a user could add support for a new compiler without writing any frontend code. Feel free to merge when you're ready!

Cheers,
Chris

compiler_gym/envs/compiler_env.py
bwasti: oh yea, I was lazy here. let me add that

compiler_gym/envs/compiler_env.py
bwasti: oh it looks like only random_search expected it. I'll fix it there.

PR #171, opened 2021-04-07
[llvm] Strip module and source IDs in benchmarks.

ChrisCummins:
This strips the ModuleID and source_filename values from the LLVM
module immediately after creation. Previously those values would leak
into the IR:

```py
>>> env = gym.make("llvm-v0")
>>> env.reset(benchmark="cBench-v1/crc32")
>>> print(env.ir)
; ModuleID = 'benchmark://cBench-v1/crc32'
source_filename = "llvm-link"
...
```

This provided an opportunity for a learning algorithm to overfit to
the benchmarks by their name. Now, the ModuleID and source_filename
attributes are stripped, anonymizing the IRs:

```py
>>> print(env.ir)
; ModuleID = '-'
source_filename = "-"
...
```


hughleat (approved):
LGTM

PR #177, opened 2021-04-08
[util] Use more granular locking for download().

ChrisCummins:
And reduce race likelihood by writing to a temporary file and renaming
it.


benoitsteiner (approved):
LGTM

PR #190, opened 2021-04-21
[datasets-overhaul #1] Add a new Benchmark class.

ChrisCummins:
A benchmark represents a particular program that is being compiled. A benchmark comprises the data that is fed into the compiler, identified by a URI.

This new Benchmark class replaces the original implementation which used only string URIs in the frontend to represent benchmarks. The new class allows additional functionality such as runtime checks or methods for validating the semantics of a benchmark to be attached to the object. 

Compiler environments may provide helper functions for generating benchmarks, such as `env.make_benchmark()` for LLVM.

The data underlying a Benchmark instance should be considered immutable. New attributes cannot be assigned to Benchmark instances.

Issue #45.

### DIFF
--- a/compiler_gym/datasets/benchmark.py
+++ b/compiler_gym/datasets/benchmark.py
@@ -1,0 +1,359 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import re
+from concurrent.futures import as_completed
+from pathlib import Path
+from typing import Callable, Iterable, List, NamedTuple, Optional, Union
+
+from compiler_gym.service.proto import Benchmark as BenchmarkProto
+from compiler_gym.service.proto import File
+from compiler_gym.util import thread_pool
+from compiler_gym.util.decorators import memoized_property
+from compiler_gym.validation_result import ValidationError
+
+# A validation callback is a function that takes a single CompilerEnv instance
+# as its argument and returns an iterable sequence of zero or more
+# ValidationError tuples.
+ValidationCallback = Callable[["CompilerEnv"], Iterable[ValidationError]]  # noqa: F821
+
+
+# Regular expression that matches the full two-part URI prefix of a dataset:
+#     {{protocol}}://{{dataset}}
+#
+# A trailing slash is permitted.
+#
+# Example matches: "benchmark://foo-v0", "benchmark://foo-v0/".
+DATASET_NAME_RE = re.compile(
+    r"(?P<dataset>(?P<dataset_protocol>[a-zA-z0-9-_]+)://(?P<dataset_name>[a-zA-z0-9-_]+-v(?P<dataset_version>[0-9]+)))/?"
+)
+
+# Regular expression that matches the full three-part format of a benchmark URI:
+#     {{protocol}}://{{dataset}}/{{id}}
+#
+# The {{id}} is optional.
+#
+# Example matches: "benchmark://foo-v0/" or "benchmark://foo-v0/program".
+BENCHMARK_URI_RE = re.compile(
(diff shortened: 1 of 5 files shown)


compiler_gym/datasets/benchmark.py
```diff
@@ -0,0 +1,350 @@
...
+# E.g. "benchmark://foo-v0".
+DATASET_NAME_RE = re.compile(
+    r"(?P<dataset>(?P<dataset_protocol>[a-zA-z0-9-_]+)://(?P<dataset_name>[a-zA-z0-9-_]+-v(?P<dataset_version>[0-9]+)))"
+)
+
+# Regular expression that matches the full three-part format of a benchmark URI:
+#     <protocol>://<dataset>/<id>
+#
+# E.g. "benchmark://foo-v0/" or "benchmark://foo-v0/program".
+BENCHMARK_URI_RE = re.compile(
```
benoitsteiner, line 38: I'd change DATASET_NAME_RE to accept both benchmark://foo-v0 and benchmark://foo-v0/, and BENCHMARK_URI_RE to require all 3 parts, i.e benchmark://foo-v0/program

benoitsteiner (approved):
LGTM.

compiler_gym/datasets/benchmark.py
```diff
@@ -0,0 +1,350 @@
...
+    """A benchmark which has a single source file."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._src_name = None
+        self._src_path = None
+
+    @classmethod
+    def create(
+        cls, uri: str, bitcode_path: Path, src_name: str, src_path: Path
```
hughleat: This is very LLVM - if there are going to be multiple compilers, should all references things like bitcode be removed?
Consider moving this to LLVMBenchmark

compiler_gym/datasets/benchmark.py
```diff
@@ -0,0 +1,350 @@
...
+    def from_file(cls, uri: str, path: Path):
+        """Construct a benchmark from the path to a file.
+
+        :param uri: The URI of the benchmark.
+
+        :param path: A filesystem path.
+
+        :return: A :class:`Benchmark <compiler_gym.datasets.Benchmark>` instance.
+        """
+        return cls(
```
hughleat, line 300: What if the path is a dir?

compiler_gym/datasets/benchmark.py
```diff
@@ -0,0 +1,350 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._src_name = None
+        self._src_path = None
+
+    @classmethod
+    def create(
+        cls, uri: str, bitcode_path: Path, src_name: str, src_path: Path
+    ) -> Benchmark:
+        """Create a benchmark from paths."""
```
hughleat, line 343: Does src_path have to be a file?
Also, can't multiple files contribute to a bitcode? Headers, etc...
Consider `sources`?

hughleat (approved):
LGTM - couple of small questions about one class.

compiler_gym/datasets/benchmark.py
ChrisCummins: Good idea, done

compiler_gym/datasets/benchmark.py
ChrisCummins: Good catch! This was initially LLVM-specific. I refactored it bust missed a couple of bitcode references. Fixed

compiler_gym/datasets/benchmark.py
ChrisCummins: Directories are not supported (could be added in the future with a `from_dir()`).

I've added a check that raises `FileNotFoundError` if `path` is not a file.

compiler_gym/datasets/benchmark.py
ChrisCummins: Yep, multiple sources are supported by passing a list of named files to `Benchmark()` constructor. This `BenchmarkWithSource()` class is just a shortcut for the common case where a benchmark has a single source.

PR #191, opened 2021-04-21
[datasets-overhaul #2] Add a new Dataset class.

ChrisCummins:
A dataset is a collection of benchmarks.

Datasets provide a convenience abstraction for installing and managing groups of benchmarks. Datasets provide the API for instantiating benchmarks, either by randomly selecting from the available benchmarks or by selecting by URI.

The Dataset class is an abstract base for implementing datasets. At a minimum, subclasses must implement the `benchmark()` and `benchmark_uris()` methods. Other methods such as `install()` may be used where helpful.

Issue #45.

### DIFF
--- a/tests/datasets/dataset_test.py
+++ b/tests/datasets/dataset_test.py
@@ -1,0 +1,138 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Unit tests for //compiler_gym/datasets."""
+from pathlib import Path
+
+import pytest
+
+from compiler_gym.datasets.dataset import Dataset
+from tests.test_main import main
+
+pytest_plugins = ["tests.pytest_plugins.common"]
+
+# pylint: disable=abstract-method
+
+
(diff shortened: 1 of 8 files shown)


tests/datasets/dataset_test.py
```diff
@@ -0,0 +1,97 @@
...
+from compiler_gym.datasets.dataset import Dataset
+from tests.test_main import main
+
+pytest_plugins = ["tests.pytest_plugins.common"]
+
+# pylint: disable=abstract-method
+
+
+def test_dataset_properties(tmpwd: Path):
+    del tmpwd  # Unused
```
JD-ETH: why is the argument required for the test function?

JD-ETH (approved):
I think the structure is very clear now this way, don't see issues here.

tests/datasets/dataset_test.py
ChrisCummins: Good catch! In an early revision, `Dataset` would create a directory on init, so the test needed a scrap working directory to write to. This is no longer required, I'll remove it.

ChrisCummins:
Thanks for the review @JD-ETH!

PR #192, opened 2021-04-21
[datasets-overhaul #3] Add Dataset helper subclasses.

ChrisCummins:
This adds three helper classes for describing datasets:

* `FilesDataset` - A dataset comprising a directory tree files.
* `TarDataset` - A dataset comprising a files tree stored in a tar archive.
* `TarDatasetWithManifest` - A tarball-based dataset that has a manifest file which lists the URIs.

Issue #45.

### DIFF
--- a/compiler_gym/datasets/files_dataset.py
+++ b/compiler_gym/datasets/files_dataset.py
@@ -1,0 +1,163 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+import os
+from pathlib import Path
+from typing import Iterable, List, Optional
+
+from compiler_gym.datasets.dataset import Benchmark, Dataset
+from compiler_gym.util.decorators import memoized_property
+
+
+class FilesDataset(Dataset):
+    """A dataset comprising a directory tree of files.
+
+    A FilesDataset is a root directory that contains (a possibly nested tree of)
+    files, where each file represents a benchmark. Files can be filtered on
+    their expected filename suffix.
+
+    Every file that matches a filename suffix is a benchmark. The URI of a
+    benchmark is the relative path of the file, stripped of the filename suffix.
+    For example, given the following file tree:
+
+    .. code-block::
+
+        /tmp/dataset/a.txt
+        /tmp/dataset/LICENSE
+        /tmp/dataset/subdir/subdir/b.txt
+        /tmp/dataset/subdir/subdir/c.txt
+
+    a FilesDataset :code:`benchmark://ds-v0` rooted at :code:`/tmp/dataset` with
+    filename suffix :code:`.txt` will contain the following URIs:
+
+        >>> list(dataset.benchmark_uris())
+        [
+            "benchamrk://ds-v0/a",
+            "benchamrk://ds-v0/subdir/subdir/b",
+            "benchamrk://ds-v0/subdir/subdir/c",
+        ]
+    """
+
+    def __init__(
+        self,
+        dataset_root: Path,
+        benchmark_file_suffix: str = "",
+        memoize_uris: bool = True,
+        **dataset_args,
+    ):
+        """Constructor.
+
+        :param dataset_root: The root directory to look for benchmark files.
+
+        :param benchmark_file_suffix: A file extension that must be matched for
+            a file to be used as a benchmark.
+
+        :param memoize_uris: Whether to memoize the list of URIs contained in
+            the dataset. Memoizing the URIs is a tradeoff between *O(n)*
+            computation complexity of random access vs *O(n)* space complexity
+            of memoizing the URI list.
+
+        :param dataset_args: See :meth:`Dataset.__init__()
+            <compiler_gym.datasets.Dataset.__init__>`.
+        """
+        super().__init__(**dataset_args)
+        self.dataset_root = dataset_root
+        self.benchmark_file_suffix = benchmark_file_suffix
+        self.memoize_uris = memoize_uris
+        self._memoized_uris = None
+
+    @memoized_property
+    def n(self) -> int:  # pylint: disable=invalid-overriden-method
+        self.install()
+        return sum(
+            sum(1 for f in files if f.endswith(self.benchmark_file_suffix))
+            for (_, _, files) in os.walk(self.dataset_root)
+        )
+
+    @property
+    def _benchmark_uris_iter(self) -> Iterable[str]:
+        """Return an iterator over benchmark URIs that is consistent across runs."""
+        self.install()
+        for root, dirs, files in os.walk(self.dataset_root):
+            # Sort the subdirectories so that os.walk() order is stable between
+            # runs.
+            dirs.sort()
+            reldir = root[len(str(self.dataset_root)) + 1 :]
+            for filename in sorted(files):
+                # If we have an expected file suffix then ignore files that do
+                # not match, and strip the suffix from files that do match.
+                if self.benchmark_file_suffix:
+                    if not filename.endswith(self.benchmark_file_suffix):
+                        continue
+                    filename = filename[: -len(self.benchmark_file_suffix)]
+                # Use os.path.join() rather than simple '/' concatenation as
+                # reldir may be empty.
+                yield os.path.join(self.name, reldir, filename)
+
+    @property
+    def _benchmark_uris(self) -> List[str]:
+        return list(self._benchmark_uris_iter)
+
+    def benchmark_uris(self) -> Iterable[str]:
+        if self._memoized_uris:
+            yield from self._memoized_uris
+        elif self.memoize_uris:
+            self._memoized_uris = self._benchmark_uris
+            yield from self._memoized_uris
+        else:
+            yield from self._benchmark_uris_iter
+
+    def benchmark(self, uri: Optional[str] = None) -> Benchmark:
+        self.install()
+        if uri is None or len(uri) <= len(self.name) + 1:
+            if not self.n:
+                raise ValueError("No benchmarks")
+            return self._get_benchmark_by_index(self.random.integers(self.n))
+
+        relpath = f"{uri[len(self.name) + 1:]}{self.benchmark_file_suffix}"
+        abspath = self.dataset_root / relpath
+        if not abspath.is_file():
+            raise LookupError(f"Benchmark not found: {uri} (file not found: {abspath})")
+        return self.benchmark_class.from_file(uri, abspath)
+
+    def _get_benchmark_by_index(self, n: int) -> Benchmark:
+        """Look up a benchmark using a numeric index into the list of URIs,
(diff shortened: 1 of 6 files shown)


compiler_gym/datasets/files_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+        return self.benchmark_class.from_file(uri, abspath)
+
+    def get_benchmark_by_index(self, n: int) -> Benchmark:
+        """Look a benchmark using a numeric index into the list of URIs."""
+        # If we have memoized the benchmark IDs then just index into the list.
+        # Otherwise we will scan through the directory hierarchy.
+        if self.memoize_uris:
+            if not self._memoized_uris:
+                self._memoized_uris = self._benchmark_uris
+            return self.benchmark(self._memoized_uris[n])
```
JD-ETH, line 133: a range check here?

compiler_gym/datasets/files_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+                # the target file index, so skip the whole lot.
+                i += len(valid_files)
+                dirs.sort()
+            else:
+                # Iterate through the files in the directory in order until we
+                # reach the target index.
+                for filename in sorted(valid_files):
+                    if i == n:
+                        name_stem = filename[: -len(self.benchmark_file_suffix)]
+                        # Use os.path.join() rather than simple '/' concaentation as
```
JD-ETH:
```suggestion
                        # Use os.path.join() rather than simple '/' concatenation as
```

compiler_gym/datasets/files_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+
+            if i + len(valid_files) < n:
+                # There aren't enough files in this directory to bring us up to
+                # the target file index, so skip the whole lot.
+                i += len(valid_files)
+                dirs.sort()
+            else:
+                # Iterate through the files in the directory in order until we
+                # reach the target index.
+                for filename in sorted(valid_files):
```
JD-ETH: Can't you access the valid_files[n-i] directly instead?

compiler_gym/datasets/files_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+            reldir = root[len(str(self.dataset_root)) + 1 :]
+
+            # Filter only the files that match the target file suffix.
+            valid_files = [f for f in files if f.endswith(self.benchmark_file_suffix)]
+
+            if i + len(valid_files) < n:
+                # There aren't enough files in this directory to bring us up to
+                # the target file index, so skip the whole lot.
+                i += len(valid_files)
+                dirs.sort()
```
JD-ETH, line 149: 1.Why sort? 
2. if there isn't enough files in the dataset_root compared to n, why not throw immediately? And why would i not be equal to 0 here?

compiler_gym/datasets/tar_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+        if not self._installed:
+            self._installed = self._tar_extracted_marker.is_file()
+        return self._installed
+
+    def install(self) -> None:
+        if self.installed:
+            return
+
+        # Thread-level and process-level locks to prevent races.
+        with self._tar_lock, fasteners.InterProcessLock(self._tar_lockfile):
```
JD-ETH: wow this is cool

compiler_gym/datasets/tar_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+    @memoized_property
+    def _benchmark_uris(self) -> List[str]:
+        """Fetch or download the URI list."""
+        if self._manifest_path.is_file():
+            return self._read_manifest_file()
+
+        with self._manifest_lock:
+            with fasteners.InterProcessLock(self._manifest_lockfile):
+                # Now that we have acquired the lock, repeat the check.
+                if self._manifest_path.is_file():
```
JD-ETH: why would this happen? So there is a case without lock the file won't be consider existing?

compiler_gym/datasets/tar_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+                manifest_data = io.BytesIO(
+                    download(self.manifest_urls, self.manifest_sha256)
+                )
+                if self.manifest_compression == "bz2":
+                    with bz2.BZ2File(manifest_data) as f:
+                        manifest_data = f.read()
+                elif self.manifest_compression == "gz":
+                    with gzip.GzipFile(fileobj=manifest_data) as f:
+                        manifest_data = f.read()
+                else:
```
JD-ETH: manifest sounds like a metadata for me, and why can't it just be plain txt file? 
I guess you have specific dataset example in mind here.

compiler_gym/datasets/tar_dataset.py
```diff
@@ -0,0 +1,204 @@
...
+                    raise TypeError(
+                        f"Unknown manifest compression: {self.manifest_compression}"
+                    )
+
+                # Write to a temporary file and rename to avoid a race.
+                with open(f"{self._manifest_path}.tmp", "wb") as f:
+                    f.write(manifest_data)
+                os.rename(f"{self._manifest_path}.tmp", self._manifest_path)
+
+                lines = manifest_data.decode("utf-8").rstrip().split("\n")
```
JD-ETH: what about `_read_manifest_file` again()?

tests/datasets/files_dataset_test.py
```diff
@@ -0,0 +1,150 @@
...
+    assert benchmarks_b != benchmarks_c
+
+    assert set(benchmarks_a) == set(populated_dataset.benchmark_uris())
+    assert set(benchmarks_c) == set(populated_dataset.benchmark_uris())
+
+
+def test_populated_dataset_get_benchmark_by_index(populated_dataset: FilesDataset):
+    for i in range(populated_dataset.n):
+        populated_dataset.get_benchmark_by_index(i)
+
```
JD-ETH, line 200: some assert seem to be missing here?

compiler_gym/datasets/files_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+            sum(1 for f in files if f.endswith(self.benchmark_file_suffix))
+            for (_, _, files) in os.walk(self.dataset_root)
+        )
+
+    @property
+    def _benchmark_uris_iter(self) -> Iterable[str]:
+        """Return an iterator over benchmark URIs is consistent across runs."""
+        self.install()
+        for root, dirs, files in os.walk(self.dataset_root):
+            dirs.sort()
```
JD-ETH, line 85: what does the sort do here?

compiler_gym/datasets/files_dataset.py
```diff
@@ -0,0 +1,156 @@
...
+                # Iterate through the files in the directory in order until we
+                # reach the target index.
+                for filename in sorted(valid_files):
+                    if i == n:
+                        name_stem = filename[: -len(self.benchmark_file_suffix)]
+                        # Use os.path.join() rather than simple '/' concaentation as
+                        # reldir may be empty.
+                        uri = os.path.join(self.name, reldir, name_stem)
+                        return self.benchmark_class.from_file(uri, f"{root}/{filename}")
+                    i += 1
```
JD-ETH: I am bit confused here. Doesn't *files* contain all files in the directory?

tests/datasets/files_dataset_test.py
JD-ETH: Also seeing some out of range throws would be nice.

JD-ETH (changes requested):

compiler_gym/datasets/files_dataset.py
ChrisCummins: It ensures that the order of this iterator is consistent across runs. [os.walk() calls os.scandir()](https://github.com/python/cpython/blob/a07da09ad5bd7d234ccd084a3a0933c290d1b592/Lib/os.py#L357), and [os.scandir() iteration order is arbitrary](https://docs.python.org/3/library/os.html#os.scandir).

ChrisCummins: I'll add a comment to clarify.

compiler_gym/datasets/files_dataset.py
ChrisCummins: `get_benchmark_by_index()` is a private method and we only call with values we know to be valid. I'll rename this method to `_get_benchmark_by_index` to make it clear that its private, and add a comment that no bounds checking is performed.

compiler_gym/datasets/files_dataset.py
ChrisCummins: 1. Sort to keep iteration order consistent (see above).
2. `os.walk()` returns a list of files in the current directory, and a list of subdirectories that will be visited next. If we don't have enough files in the current directory, then we can descend into the subdirectories.

compiler_gym/datasets/files_dataset.py
ChrisCummins: Good catch :) Thanks!

compiler_gym/datasets/files_dataset.py
ChrisCummins: 👍

compiler_gym/datasets/files_dataset.py
ChrisCummins: You're right, this loop wasn't required. Removed

compiler_gym/datasets/tar_dataset.py
ChrisCummins: In general, manifests mean "lists of stuff", but it is an overloaded term. In bazel, a manifest is a list of file paths, one per line. In python packaging, a manifest is a file that has its own special syntax and commands. I'm using the term "manifest" here the same way that bazel does. I'll add an explanation in the docstring to make it clear.

compiler_gym/datasets/tar_dataset.py
ChrisCummins: Consider two threads *X* and *Y* that both call into this function. *X* gets the lock first, causing *Y* to wait. *X* downloads the file and releases the lock. *Y*, having acquired the lock, should now check to see whether the file needs downloading, as otherwise it would be redundantly repeating the work that *X* just completed.

Hope that makes sense :)

compiler_gym/datasets/tar_dataset.py
ChrisCummins: Good catch. We already have `manifest_data` in-memory, so we can skip the disk op. But there is a bit of redundant logic here. I'll refactor it out :)

tests/datasets/files_dataset_test.py
ChrisCummins: Ahah yes looks like I forgot to add any tests :D Done

ChrisCummins:
Thanks for the very detailed review @JD-ETH! I have gone through and updated the PR. Lemme know if that LGTY.

Cheers,
Chris

PR #193, opened 2021-04-22
[datasets-overhaul #4] Add a Datasets class for managing datasets.

ChrisCummins:
This `Datasets` provides a dictionary-like interface for indexing and iterating
over multiple `Dataset` objects. Select a dataset by name using:

```py
>>> env.datasets["benchmark://cbench-v1"]
```

Check whether a dataset exists using:

```py
>>> "benchmark://cbench-v1" in env.datasets
True
```

Or iterate over the collection using:

```py
>>> for dataset in env.datasets:
...     print(dataset.name)
benchmark://cbench-v1
benchmark://github-v0
benchmark://npb-v0
```

To select a benchmark from the datasets, use `benchmark()`:

```py
>>> env.datasets.benchmark("benchmark://a-v0/a")
```

Use the `benchmarks()` method to iterate over every benchmark in the
datasets in a stable round robin order:

```py
>>> for benchmark in env.datasets.benchmarks():
...     print(benchmark)
benchmark://cbench-v1/1
benchmark://github-v0/1
benchmark://npb-v0/1
benchmark://cbench-v1/2
...
```

If you want to exclude a dataset, delete it:

```py
>>> del env.datasets["benchmark://b-v0"]
```

To iterate over the benchmarks in a random order, use `benchmark()` and omit the URI:

```py
>>> for i in range(100):
...     benchmark = env.datasets.benchmark()
```

This uses uniform random selection to sample across datasets. For finite
datasets, you could weight the sample by the size of each dataset:

```py
>>> weights = [len(d) for d in env.datasets]
>>> np.random.choice(list(env.datasets), p=weights).benchmark()
```


compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+    The datasets object can be also be iterated over to list the available
+    datasets:
+
+        >>> for dataset in env.datasets:
+        ...     print(dataset.name)
+        benchmark://cbench-v1
+        benchmark://github-v0
+        benchmark://npb-v0
+
+    For the second goal, the datasets interface adds an *active* vs *installed*
```
hughleat: IMO this is a weird thing to have. I'm not convinced I understand the use case.
Why not have everything be installed and then layer something over the top as a separate class - for the few use cases that might need this.
E.g. if I want to do something to half the benchmarks I have access to write a filter.

I'd suggest dropping the second goal.

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+    def __init__(
+        self,
+        datasets: Iterable[Dataset],
+        random: Optional[np.random.Generator] = None,
+    ):
+        # A mapping from dataset name to the dataset instance.
+        self._datasets: Dict[str, Dataset] = {d.name: d for d in datasets}
+        # The set of activate datasets. All datasets except hidden ones are
+        # initialized as activate.
+        self._active_datasets: Set[str] = set(
```
hughleat: I don't get the need for this.

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+            dataset.seed(seed)
+
+    def datasets(
+        self, inactive: bool = False, hidden: bool = False
+    ) -> Iterable[Dataset]:
+        """Enumerate the datasets.
+
+        Dataset order is consistent across runs.
+
+        :param inactive: If :code:`False` (the default), only datasets that are
```
hughleat: Why is this needed?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+            active will be returned.
+
+        :param hidden: If :code:`False` (the default), only datasets whose
```
hughleat, line 125: Why is this needed?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+            :meth:`Dataset.hidden <compiler_gym.datasets.Dataset.hidden>` value
```
hughleat, line 126: Can't they have an `is_installed` method on the dataset (or possibly benchmark)?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+        If a name is given, return the corresponding :meth:`Dataset
+        <compiler_gym.datasets.Dataset>`. Else, return a dataset uniformly
+        randomly from the set of available datasets.
+
+        Use :meth:`seed() <compiler_gym.datasets.Dataset.seed>` to force a
+        reproducible order for randomly selected datasets.
+
+        Name lookup will succeed whether or not the dataset is active or hidden.
+
+        :param name: A dataset name, or :code:`None` to select a dataset
```
hughleat: Does not match prototype

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+        :raises LookupError: If :code:`name` is not found.
+        """
+        if dataset is None:
+            if not self._active_datasets:
+                raise ValueError("No active datasets available")
+
+            return self._datasets[self.random.choice(list(self._active_datasets))]
+
+        if isinstance(dataset, Benchmark):
```
hughleat: Does not seem to match typing in prototype

hughleat: Also, not 100% sure I understand this.
What's the purpose? Why not call something like:
`benchmark.get_dataset() -> Optional[DataSet]` (Optional in case you have orphan benchmarks)

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+            dataset_name = dataset.name
+        else:
+            dataset_name = resolve_uri_protocol(dataset)
+
+        if dataset_name not in self._datasets:
+            raise LookupError(f"Dataset not found: {dataset_name}")
+        return self._datasets[dataset_name]
+
+    def __getitem__(self, name: str) -> Dataset:
```
hughleat: Why doesn't this match the prototype of `dataset()`?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+        added = True
+        if dataset.name in self._datasets:
+            added = False
+            warnings.warn(f"Replacing existing dataset: {dataset.name}")
+        self._datasets[dataset.name] = dataset
+        if not dataset.hidden:
+            self._active_datasets.add(dataset.name)
+        return added
+
+    def remove(self, dataset: Union[str, Dataset]) -> bool:
```
hughleat: Why another odd way of hiding datasets? Why not just uninstall?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def install(self, dataset: Union[str, Dataset]) -> bool:
+        """Install a dataset.
+
+        Run the local installation procedure for the given dataset. This will
+        perform one-time setup of the dataset such as downloading remote
+        resources. See :meth:`Dataset.install()
+        <compiler_gym.datasets.Dataset.install>`.
+
+        Installing a dataset does not make it inactive if it is not already.
```
hughleat: Do you mean 'active'?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+        :param dataset: A :meth:`Dataset <compiler_gym.datasets.Dataset>`
+            instance, or the name of a dataset.
+
+        :return: :code:`True` if the dataset was installed, :code:`False` if
+            it was already installed.
+        """
+        dataset: Dataset = self.dataset(dataset)
+        installed = not dataset.installed
+        dataset.install()
```
hughleat: Why is this called again if already installed?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+        :return: :code:`True` if the dataset was uninstalled, :code:`False` if
+            it was already uninstalled.
+        """
+        dataset: Dataset = self.dataset(dataset)
+        uninstalled = dataset.installed
+        dataset.uninstall()
+        return uninstalled
+
+    def require(self, dataset: Union[str, Dataset]) -> bool:
```
hughleat: IMO Way too much stuff here. Just install or uninstall.

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+        if hasattr(dataset, "name"):
+            added = self.add(dataset)
+            dataset_name = dataset.name
+        else:
+            dataset_name = resolve_uri_protocol(dataset)
+        activated = self.activate(dataset_name)
+        installed = self.install(dataset_name)
+        return added or activated or installed
+
+    def benchmarks(self) -> Iterable[Benchmark]:
```
hughleat, line 229: Presumably datasets is finite?
If the first dataset is infinite, then you will never get any benchmarks from the second.
Consider:
```
iters = [d.benchmarks().iterator() for d in self.datasets()]
while iters:
  it = iters.pop_front()
  if it.has_next(): # I don't know how Python does iterators
    yield it.next()
    iters.push_back(it)
```
That way they get a benchmark from each suite until exhaustion and an infinite suite can't hog the limelight.

(There are better ways to achieve what I wrote above).

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+        Benchmark URI order is consistent across runs. URIs from each dataset
+        are returned in order. The order of :meth:`benchmarks()
+        <compiler_gym.datasets.Datasets.benchmarks>` and :meth:`benchmark_uris()
+        <compiler_gym.datasets.Datasets.benchmark_uris>` is the same.
+
+        :return: An iterable sequence of benchmark URI strings.
+        """
+        for dataset in self.datasets():
+            yield from dataset.benchmark_uris()
```
hughleat: Same deal about infinite suite stopping any others getting a go as prev method

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+    def benchmark(self, uri: Optional[str] = None) -> Benchmark:
+        """Select a benchmark.
+
+        If a benchmark URI is given, the corresponding :class:`Benchamrk
```
hughleat: Benchamrk -> Benchmark

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+        <compiler_gym.datasets.Benchmark>` is returned, regardless of whether
+        the containing dataset is active, hidden, or installed.
+
+        If no URI is given, a benchmark is selected randomly. Fist, a dataset is
```
hughleat: Fist -> First

hughleat (commented):
Some thoughts :-)

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+        smaller datasets will be overrepresented and large datasets will be
+        underrepresented.
+
+        Use :meth:`seed() <compiler_gym.datasets.Dataset.seed>` to force a
+        reproducible order for randomly selected benchmarks.
+
+        :param uri: The URI of the benchmark to return. If :code:`None`, select
+            a benchmark randomly using :code:`self.random`.
+
+        :return: A :class:`Benchamrk <compiler_gym.datasets.Benchmark>`
```
benoitsteiner: s/Benchamrk/Benchmark/

ChrisCummins:
Hey @hughleat, thanks for taking a look! Lemme have a go at explaining this one, as I think its most important:

> Why not have everything be installed and then layer something over the top as a separate class - for the few use cases that might need this.

That's pretty much what I'm aiming for here. Everything is installed, and `Datasets` is a layer over the top that allows the user to control what is "available" to the environment. It all boils down to controlling what gets returned by the "give me a random benchmark" function. E.g. consider the user who wants to run something over a random 100 benchmarks:

```py
for _ in range(100):
    bm = env.datasets.benchmark()
    do_a_thing(env, bm)
``` 

The idea of the `Datasets` class is that it makes it easy for the user to control precisely which datasets are used to draw upon for selecting random benchmarks. E.g.

```py
# Setup environment to use just cbench-v0 benchmarks
for dataset in env.datasets:
    env.deactivate(dataset)
env.require("cbench-v0")
```

This is entirely decoupled from the actual `install()` and `uninstall()` operations which do the actual file wrangling.

Perhaps the current approach is overcomplicated. The simplest alternative would be just to use a dictionary of datasets like:

```py
env.datasets = {
  "cbench-v1": CBenchDataset(...),
  "npb-v0": NpbDataset(...),
}
```

the user could roll their own random benchmark selector, using:

```py
for _ in range(100):
    bm = random.choice(list(env.datasets.values())).benchmark()
    do_a_thing(env, bm)
```

Lemme know what you think.

Cheers,
Chris

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,480 @@
...
+
+        :return: A :meth:`Dataset <compiler_gym.datasets.Dataset>` instance.
+
+        :raises LookupError: If :code:`name` is not found.
+        """
+        if dataset is None:
+            if not self._active_datasets:
+                raise ValueError("No active datasets available")
+
+            return self._datasets[self.random.choice(list(self._active_datasets))]
```
bwasti: is there a performance (or scaling) reason behind encapsulating the random selections?

users might be comfortable with randomly choosing the datasets directly

e.g.
```
random.choice(env.datasets)
```
or
```
env.dataset(idx=random.randint(0, len(env.datasets)))
```

ChrisCummins:
Ooh and one thing I forgot, the idea of a "hidden" dataset serves one very particular purpose: to allow us to deprecate datasets without immediately breaking all user code. E.g. we can make `env.datasets["cBench-v0"].benchmark()` emit a deprecation warning requesting that users switch to "cbench-v1", without "cBench-v0" showing up anywhere else unless explicitly asked for.

Cheers,
Chris

bwasti (commented):
Is the activate/deactivate paradigm to avoid the pain of re-installations? If so, I think a one time hit for first `install` (with warnings + nice tqdm bar + cache for reinstallation) would be fine to most users.  (Although you'd need `force_uninstall` for clearing the cache)

ChrisCummins:
> Is the activate/deactivate paradigm to avoid the pain of re-installations?

Yep, exactly that. `activate()` / `deactivate()` are just ways of controlling what benchmarks the environment has access to. I think Hugh is right and that this can be simplified.

`install()` and `uninstall()` are where the actual work gets down. `install()` isn't just about downloading pre-packaged programs - in the case of [csmith-v0](https://github.com/facebookresearch/CompilerGym/blob/635be65414d48bad2c37cdd1813da8f7631d05c1/compiler_gym/envs/llvm/datasets/csmith.py#L130-L140) I use `install()` to download and compile the program generator. `uninstall()` is for cleaning up / freeing disk space.

Cheers,
Chris

compiler_gym/datasets/datasets.py
ChrisCummins: My thinking was only that calling `Dataset.benchmark()` without a URI will return a random benchmark, so `Datasets.benchmark()` should too. There's no particular reason to encapsulate it, and the docs should be clear about the equivalence to the `random.choice(list(env.dataset)).benchmark()`.

ChrisCummins: Being more explicit would let you do more fancy things too, like selecting a random benchmark weighted by dataset size:

```py
np.random.choice(list(env.datasets), p=[d.n for d in env.datasets]).benchmark()
```

hughleat:
> The idea of the `Datasets` class is that it makes it easy for the user to control precisely which datasets are used to draw upon for selecting random benchmarks. E.g.

I guess I wonder how common that use case is. They have to be particular about which suite a random benchmark comes from - but only in the case where it comes from more than one, because they have an easy way to select from one. They can't be particular about which benchmarks in those 2+ suites. And they aren't particular about the distribution.  Is this really a problem that needs to be solved for them anyway. If they are anywhere just off that, then they'll need to do something else anyway.

IMO, ditch the whole active/inactive thing and don't replace it with anything.

ChrisCummins:
> IMO, ditch the whole active/inactive thing and don't replace it with anything.

Agreed, thanks for making the case for simplifying the API, I think that would improve things.

For v2 of this PR I'll draft a commit which drops the following methods:

* `Datasets.activate()`
* `Datasets.deactivate()`
* `Datasets.is_active()`
* `Datasets.install()` -> no need, equivalent to `datasets[...].install()`
* `Datasets.uninstall()` -> no need, equivalent to `datasets[...].uninstall()`
* `Datasets.add()` -> replace with `Datasets.__setitem__()`, see below
* `Datasets.remove()` -> replace with `Datasets.__delitem__()`, see below

In addition the `__getitem__()` and `__iteritem()`, a couple of other operator overloads could be useful for making this class behave more like a regular python dict:

```py
del env.datasets["cbench-v0"]  # remove a dataset
env.datasets["testset-v0"] = ds  # add a dataset
len(datasets)   # num datasets
```

Cheers,
Chris

compiler_gym/datasets/datasets.py
ChrisCummins: v2 rewrite: Renamed to `_visible_datasets`. I still need to keep track of the subset of `self._datasets` that can be used for iterating over

compiler_gym/datasets/datasets.py
ChrisCummins: Removed

compiler_gym/datasets/datasets.py
ChrisCummins: Hidden datasets are used so that we can deprecate things gracefully. E.g.

```py
>>> list(env.datasets.dataset())
benchmark://npb-v0
benchmark://cbench-v2
>>> list(env.datasets.dataset(hidden=True))
benchmark://npb-v0
benchmark://cbench-v0
benchmark://cbench-v1
benchmark://cbench-v2
>>> env.datasets.benchmark("benchmark://cbench-v0/crc32")
# works, but will raise a deprecation warning
```

compiler_gym/datasets/datasets.py
ChrisCummins: Yep. I've removed the `instal()` and `uninstall()` methods from `Datasets`.

compiler_gym/datasets/datasets.py
ChrisCummins: I like this a lot!

Python iterators raise `StopIteration` when empty, so its:

```py
def round_robin_iterables(iters: Iterable[Iterable[T]]) -> Iterable[T]:
    """Yield from the given iterators in round robin order."""
    iters = deque(iters)
    while iters:
        it = iters.popleft()
        try:
            yield next(it)
            iters.append(it)
        except StopIteration:
            pass
```

compiler_gym/datasets/datasets.py
ChrisCummins: Done

compiler_gym/datasets/datasets.py
ChrisCummins: Thanks! Turns out I made that typo all over the place :) #199

compiler_gym/datasets/datasets.py
ChrisCummins: Thanks! Turns out I made that typo all over the place :) #199

compiler_gym/datasets/datasets.py
ChrisCummins: Thanks, fixed

compiler_gym/datasets/datasets.py
ChrisCummins: Thanks, fixed

compiler_gym/datasets/datasets.py
ChrisCummins: Implemented

ChrisCummins:
I've updated the code and PR description with a much stripped-back interface. `Datasets` is now a "dict with benefits" :-)

Lemme know what you think.

ChrisCummins:
Eugh, managed to kill this PR by accident and GitHub won't let me reopen it. Reopened as #200.

compiler_gym/datasets/datasets.py
hughleat: Ah, I like that purpose :-)
<pedant-warning>
Consider renaming it to `include_deprecated`. First deprecated makes the purpose of the hidding clear - and will suggest to people that they should not rely on it sticking around. Second with just the name being `hidden`, I would guess it would return the hidden benchmarks, not the unhidden ones.

compiler_gym/datasets/datasets.py
hughleat: Should this also have an `include_deprecated` flag?

compiler_gym/datasets/datasets.py
```diff
@@ -0,0 +1,305 @@
...
+        fully enumerated. The order of :meth:`benchmarks()
+        <compiler_gym.datasets.Datasets.benchmarks>` and :meth:`benchmark_uris()
+        <compiler_gym.datasets.Datasets.benchmark_uris>` is the same.
+
+        :return: An iterable sequence of :class:`Benchmark
+            <compiler_gym.datasets.Benchmark>` instances.
+        """
+        return round_robin_iterables((d.benchmarks() for d in self.datasets()))
+
+    def benchmark_uris(self) -> Iterable[str]:
```
hughleat, line 243: Should this also have an include_deprecated flag?

hughleat: And, then does that suggest that each dataset probably needs it, too?

compiler_gym/datasets/datasets.py
ChrisCummins: Good idea, will update #200 with that

compiler_gym/datasets/datasets.py
ChrisCummins: I don't know if its an important use case, but sure why not :) done

compiler_gym/datasets/datasets.py
ChrisCummins: > And, then does that suggest that each dataset probably needs it, too?

Only entire datasets are deprecated or not, so I don't think we need to propagate the `with_deprecated` to them.

PR #195, opened 2021-04-22
[datasets-overhaul #1.5] Update make_benchmark() to use the new Benchmark class.

ChrisCummins:
* Change `make_benchmark()` so that it returns a `Benchmark` class instance, rather than a protocol buffer.
* Add support for assembling `*.ll` files when constructing benchmarks.

Issue #45.


benoitsteiner (approved):
LGTM

PR #198, opened 2021-04-23
[datasets] Add Dataset __getitem__, __iter__, and __len__.

ChrisCummins:
This adds python operator overloads that alias to existing methods to
make the Dataset class "feel" more like a regular python dictionary:

```py
>>> len(dataset)  # equivalent to dataset.n
23

>>> for benchmark in dataset:  # iterate over the class directly
...     pass

>>> dataset["cbench-v1/crc32"]  # key a benchmark
```

This also renames Dataset.n to Dataset.size for consistent with other
containers like np.ndarray, and returns math.inf if the number of
benchmarks is infinite, not a negative integer. The advantage of
math.inf is that will poison any integer arithemtic, e.g.

```py
>>> sum(d.size for d in datasets)
inf
```

if any one of the datasets has an infinite size. With a negative
number, this would instead compute a regular integer value.

Issue #45.

### DIFF
--- a/compiler_gym/datasets/dataset.py
+++ b/compiler_gym/datasets/dataset.py
@@ -31,9 +31,10 @@ class Dataset(object):
     The Dataset class is an abstract base for implementing datasets. At a
     minimum, subclasses must implement the :meth:`benchmark()
     <compiler_gym.datasets.Dataset.benchmark>` and :meth:`benchmark_uris()
-    <compiler_gym.datasets.Dataset.benchmark_uris>` methods. Other methods such
-    as :meth:`install() <compiler_gym.datasets.Dataset.install>` may be used
-    where helpful.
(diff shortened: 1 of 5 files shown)


JD-ETH (approved):
LGTM. My only question is what happens when doing a for loop in an infinite dataset? Does it become a infinite loop? How is this typically handled in other case?

ChrisCummins:
Thanks @JD-ETH.

> what happens when doing a for loop in an infinite dataset? Does it become a infinite loop? How is this typically handled in other case?

Yep, it will loop indefinitely. This feels like the right behavior to me, but I'll update the comments to make it super clear. Infinite `for` loops are used in the standard library, e.g. [itertools.cycle()](https://docs.python.org/3/library/itertools.html#itertools.cycle).

PR #200, opened 2021-04-23
[datasets] Add a Datasets class for managing datasets.

ChrisCummins:
This `Datasets` provides a dictionary-like interface for indexing and iterating
over multiple `Dataset` objects. Select a dataset by name using:

```py
>>> env.datasets["benchmark://cbench-v1"]
```

Check whether a dataset exists using:

```py
>>> "benchmark://cbench-v1" in env.datasets
True
```

Or iterate over the collection using:

```py
>>> for dataset in env.datasets:
...     print(dataset.name)
benchmark://cbench-v1
benchmark://github-v0
benchmark://npb-v0
```

To select a benchmark from the datasets, use `benchmark()`:

```py
>>> env.datasets.benchmark("benchmark://a-v0/a")
```

Use the `benchmarks()` method to iterate over every benchmark in the
datasets in a stable round robin order:

```py
>>> for benchmark in env.datasets.benchmarks():
...     print(benchmark)
benchmark://cbench-v1/1
benchmark://github-v0/1
benchmark://npb-v0/1
benchmark://cbench-v1/2
...
```

If you want to exclude a dataset, delete it:

```py
>>> del env.datasets["benchmark://b-v0"]
```

To iterate over the benchmarks in a random order, use `benchmark()` and omit the URI:

```py
>>> for i in range(100):
...     benchmark = env.datasets.benchmark()
```

This uses uniform random selection to sample across datasets. For finite
datasets, you could weight the sample by the size of each dataset:

```py
>>> weights = [len(d) for d in env.datasets]
>>> np.random.choice(list(env.datasets), p=weights).benchmark()
```


ChrisCummins:
Updated with @hughleat's comments from #193.

hughleat (approved):
LGTM

PR #202, opened 2021-04-23
Implement CompilerEnv.seed()

ChrisCummins:
The new Datasets API is more explicit about its source of randomness when choosing benchmarks, so implement the standard gym `env.seed()` method to allow users to set and record this seed.

One quick of the `Env.seed` method is that when called with no arguments it should choose a seed randomly and return it. Given that we do not want to rely on any RNGs (that users may also have seeded), I chose instead to use the walltime to generate a seed.

Issue #45.

### DIFF
--- a/tests/compiler_env_test.py
+++ b/tests/compiler_env_test.py
@@ -80,5 +80,21 @@ def test_logger_forced():
         env_b.close()
 
 
+def test_set_seed(env: CompilerEnv):
+    assert env.seed(1) == [1]
+    assert env.seed(2) == [2]
+
+
+def test_choose_random_seed_different_values(env: CompilerEnv):
+    a, b = env.seed(), env.seed()
+    assert a != b, f"{a} != {b}"
+
+
+def test_choose_random_seed_unique_values(env: CompilerEnv):
+    """Check that calling seed multiple times returns unique values."""
+    seeds = [env.seed()[0] for _ in range(10)]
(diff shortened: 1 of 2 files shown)


tests/compiler_env_test.py
```diff
@@ -80,5 +80,16 @@ def test_logger_forced():
...
 
+def test_set_seed(env: CompilerEnv):
+    assert env.seed(1) == [1]
+    assert env.seed(2) == [2]
+
+
+def test_choose_random_seed(env: CompilerEnv):
+    # Check that at least once a unique seed is generated.
+    seeds = {env.seed()[0] for _ in range(10)}
+    assert len(seeds) > 1
```
benoitsteiner: You should call seed twice and make sure the seeds are different. I had problems in the past since the time() function often returned the same value (may only have had seconds resolution)

benoitsteiner (changes requested):

tests/compiler_env_test.py
ChrisCummins: Good catch, done.

I was never quite sure how precise `time()` was. From what I've read it is "usually" microseconds, except on Windows where it's second resolution. `time_ns()` in Py 3.7 is a nice addition

benoitsteiner (approved):
LGTM

ChrisCummins:
Made redundant by 6f716619a11f5eed92bf18e4fa1c56056c8a42e9.

PR #204, opened 2021-04-23
[env] Add a SessionNotFound error.

ChrisCummins:
A compiler service supports multiple simultaneous sessions. Frontend
code uses a Session ID value to identify the correct session when
making requests. This session ID lookup could fail for a number of
reasons - perhaps the service has restarted, the session was evicted
due to operating constraints, or the environment was configured
incorrectly.

This patch adds anew SessionNotFound error that can be used by
services to identify this particular error case so that it can be
handled appropriately. Before, it would just raise a ValueError
because of the INVALID_ARGUMENT gRPC status code.

### DIFF
--- a/compiler_gym/envs/compiler_env.py
+++ b/compiler_gym/envs/compiler_env.py
@@ -28,6 +28,7 @@ from compiler_gym.service import (
     ServiceError,
     ServiceOSError,
     ServiceTransportError,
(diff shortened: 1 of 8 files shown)


benoitsteiner (approved):
Nice. Always good to have detailed error messages to help figure out what goes on when things go wrong.

PR #207, opened 2021-04-23
[llvm] Add Csmith and llvm-stress benchmark generators.

ChrisCummins:
This adds two new datasets, csmith-v0 and llvm-stress-v0, that are
parametrized program generators. csmith-v0 uses Csmith to generate C99
programs that are then lowered to bitcode. llvm-stress-v0 generates
random LLVM-IR.

Both generators were developed to stress test compilers, so they have
an above-average chance that a generated benchmark will cause the
compiler to enter an unexpected state.

Issue #45.


compiler_gym/envs/llvm/datasets/csmith.py
```diff
@@ -0,0 +1,268 @@
...
+    @property
+    def size(self) -> int:
+        # Actually 2^32 - 1, but practically infinite for all intents and
+        # purposes.
+        return float("inf")
+
+    def benchmark_uris(self) -> Iterable[str]:
+        return (f"{self.name}/{i}" for i in range(UINT_MAX))
+
+    def benchmark(self, uri: Optional[str] = None) -> CsmithBenchmark:
```
hughleat: Consider allowing an int32 here as well as the uri

hughleat (approved):
LGTM

compiler_gym/envs/llvm/datasets/csmith.py
ChrisCummins: Good idea, thanks. Added a separate `benchmark_from_seed()` method to CsmithDataset and LlvmStressDataset to keep the signature of `benchmark()` consistent with the other datasets

PR #211, opened 2021-04-23
[llvm] Port the cBench dataset to the new API.

ChrisCummins:
This adds the new Dataset implementation of the cBench dataset. The
validation logic isn't super tidy and could be cleaned up a bit, its
just copied over from //compiler_gym/envs/llvm:legacy_datasets.

Issue #45.


hughleat (approved):
LGTM

Issue #216, opened 2021-04-27
Use pydantic for JSON-serializable data structures.


ChrisCummins:
## 🚀 Feature

Use [pydantic](https://pydantic-docs.helpmanual.io/)'s model hierachy to represent the following data-only classes:

- [x] `CompilerEnvState`
- [x] `ValidationError`
- [x] `ValidationResult`
- [x] `ConnectionOpts`

## Motivation

Currently data classes are implemented by subclassing `NamedTuple` and writing `from_json()` / `to_json()` conversion methods. The parsing and validation logic is ad-hoc, fragile, and a maintenance burden. The problems that they solve are exactly what pydantic is intended for.

## Additional context

See also #215.

ChrisCummins:
Done.

PR #217, opened 2021-04-27
Use pydantic for CompilerEnvState.

ChrisCummins:
This changes the CompilerEnvState class to be a subclass of
[pydantic.BaseModel](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage) rather than `NamedTuple`. The reasons for this are:

1) pydantic has robust APIs for parsing and serializing/deserializing
from various data types.

2) pydantic makes it easy to add validation logic to fields. For
CompilerEnvState, the constructor will now validate that the given URI
matches the expected regex, and that the walltime is nonnegative.

Issue #216.


hughleat (approved):
LGTM

Issue #254, opened 2021-05-10
Separate the implementation of RPC service and compilation session


ChrisCummins:
## 🚀 Feature

Make it easier for users to add support for new compilers by providing base classes that allow users to "fill in the blanks" for interacting with their compiler, without having to implement all of the other RPC interface boilerplate.

## Motivation

Top add support for a new compiler, a user must write a compiler service that performs two key functions:

1. **RPC Server:** Implements a server that listens on a port for RPC requests that are then dispatched to compilation sessions.
1. **Compilation session:** Implements the compilation session interface to do the actual task of applying actions and computing observations.

This division of roles is obvious in the example compiler services. In the python example, the RPC server is implemented [here](https://github.com/facebookresearch/CompilerGym/blob/9bf8a72c9737b67045ec6fd8c73df962a1ae395b/examples/example_compiler_gym_service/service_py/example_service.py#L126-L202) and the compilation session [here](https://github.com/facebookresearch/CompilerGym/blob/9bf8a72c9737b67045ec6fd8c73df962a1ae395b/examples/example_compiler_gym_service/service_py/example_service.py#L89-L123). The compilation session is where the actual interesting work gets done. The RPC server is boilerplate, and as can be seen in the example, is not an insubstantial amount of code (and must be implemented carefully to avoid concurrency bugs / performance issues). Given that the behavior of the RPC Server is the same across all compiler services, we should provide a builtin implementation so that the user can focus only on implementing the compilation session.

## Pitch

Provide helper code for C++ and Python that makes it easier to add support for new compilers.

### Python

Add a base class that represents a single interactive session with a compiler:

```py
class CompilationSession:
    """Base class that represents an interactive compilation session."""

    compiler_version: str = ""  # compiler version
    action_spaces: List[ActionSpace] = []  # what your compiler can do
    observation_spaces: List[ObservationSpace] = []  # what features you provide

    def __init__(self, working_dir: Path, action_space_index: int, benchmark: Benchmark):
        """start a new compilation session with the given action space & benchmark"""

    def apply_action(self, action_index: int) -> Tuple[bool, Optional[ActionSpace], bool]:
        """apply an action. Returns a tuple: (end_of_session, new_action_space, action_had_no_effect).
        """

    def get_observation(self, observation_space_index: int) -> Observation:
        """compute an observation"""

    def fork(self) -> CompilationSession:
        """optional. Create a copy of current session state"""
```

[...]

PR #261, opened 2021-05-11
[service] Add a CompilationSession abstract base class.

ChrisCummins:
The CompilationSession class encapsulates an incremental compilation
session.

This is the first in a series of patches that separates the RPC server
from the compilation session.

Issue #254.

### DIFF
--- a/compiler_gym/service/compilation_session.py
+++ b/compiler_gym/service/compilation_session.py
@@ -1,0 +1,90 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+from pathlib import Path
+from typing import List, Optional, Tuple
+
+from compiler_gym.service.proto import (
+    Action,
+    ActionSpace,
+    Benchmark,
+    Observation,
+    ObservationSpace,
+)
+
+
+class CompilationSession:
+    """Base class for encapsulating an incremental compilation session.
+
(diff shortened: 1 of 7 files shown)


ChrisCummins:
Marking this ready for review but no rush to merge it while we're still discussing the design in #254.

compiler_gym/service/compilation_session.py
```diff
@@ -0,0 +1,90 @@
...
+
+    action_spaces: List[ActionSpace] = []
+    """A list of action spaces describing the capabilities of the compiler."""
+
+    observation_spaces: List[ObservationSpace] = []
+    """A list of feature vectors that this compiler provides."""
+
+    def __init__(
+        self, working_dir: Path, action_space: ActionSpace, benchmark: Benchmark
+    ):
```
hughleat, line 49: Just curious: why doesn't this return a `status` like the c++ code?

hughleat (approved):
If the protobuf stuff is okay to keep then this LGTM :-)

compiler_gym/service/compilation_session.py
ChrisCummins: We [don't use exceptions in C++](https://google.github.io/styleguide/cppguide.html#Exceptions) (FWIW I don't have particularly strong opinions on the matter), the `CompilationSession` constructor in C++ is a safe default, and any of the three methods that can fail (init, compute feature, run action) return `Status`.

For Python, exceptions are fine and all of the runtime logic that interacts with `CompilationSession` will catch exceptions and set the relevant RPC error code

Issue #252, opened 2021-05-05
Add multiprocessing support to compiler_gym.leaderboard helpers


ChrisCummins:
## 🚀 Feature

Currently the LLVM leaderboard helper method takes a callback function and evaluates it in a loop over all programs in the test set. For single-threaded agents, it would be useful if this could be parallelized to maximize throughput.

ChrisCummins:
@yuandong-tian looked into this but it turns out to be harder than expected given the current implementation. Closing a "wont fix" for now.

PR #262, opened 2021-05-11
Add multiprocessing support for `lvm_instcount.py`

yuandong-tian:
Related issue: https://github.com/facebookresearch/CompilerGym/issues/252

### DIFF
--- a/compiler_gym/leaderboard/llvm_instcount.py
+++ b/compiler_gym/leaderboard/llvm_instcount.py
@@ -31,7 +31,10 @@ import logging
 import os
 from itertools import islice
 from pathlib import Path
+import queue
 from threading import Thread
+import multiprocessing as mp
+from copy import deepcopy
 from time import sleep
 from typing import Callable, List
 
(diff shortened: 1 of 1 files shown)


compiler_gym/leaderboard/llvm_instcount.py
```diff
@@ -84,61 +87,69 @@
 # interacts with that environment with the goal of maximising cumulative reward.
 Policy = Callable[[LlvmEnv], None]
 
+def eval_policy(d):
```
ChrisCummins, line 90:
```suggestion
def _eval_policy(d):
```

Nitpick: use a private function so that the auto generate [sphinx docs](https://facebookresearch.github.io/CompilerGym/compiler_gym/leaderboard.html#module-compiler_gym.leaderboard.llvm_instcount) don't create an entry for it

compiler_gym/leaderboard/llvm_instcount.py
```diff
@@ -84,61 +87,69 @@
...
+        # Override walltime in the generated state.
+        state = env.state.copy()
+        state.walltime = timer.time
+        env.close()
+        return benchmark, state
+    except:
+        # If the code fails, then return None and we restart the same job again.
+        return benchmark, None
+
+class _EvalPolicyWorker(mp.Process):
```
ChrisCummins, line 122: I don't know if the change from `Thread` to `mp.Process` is needed here

compiler_gym/leaderboard/llvm_instcount.py
```diff
@@ -84,61 +87,69 @@
...
+
+        # print("Finish one job!", flush=True)
+        # Override walltime in the generated state.
+        state = env.state.copy()
+        state.walltime = timer.time
+        env.close()
+        return benchmark, state
+    except:
+        # If the code fails, then return None and we restart the same job again.
+        return benchmark, None
```
ChrisCummins, line 120: I would remove this bare `except` and let the exception propagate. Is there a particular type of error you're interested in?

compiler_gym/leaderboard/llvm_instcount.py
```diff
@@ -84,61 +87,69 @@
...
+            while len(curr_benchmarks) > 0:
+                args = [ (bm, self.policy) for bm in curr_benchmarks ]
+                unsolved_benchmarks = []
+                for benchmark, state in pool.imap_unordered(eval_policy, args):
+                    if state is not None:
+                        self.q.put(state)
+                    else:
+                        # The job failed, put it to the next batch. 
+                        unsolved_benchmarks.append(benchmark)
+                curr_benchmarks = unsolved_benchmarks
```
ChrisCummins, line 150: Here's where I would add an alternate path uses a regular `for` loop for when `num_process == 1`. If we drop the retry loop too, perhaps something like:

```suggestion
        if self.num_process > 1:
            with mp.Pool(self.num_process) as pool:
                args = [(bm, self.policy) for bm in self.benchmarks]
                for state in pool.imap_unordered(eval_policy, args):
                    self.q.put(state)
        else:
            for benchmark in self.benchmarks:
                state = eval_policy((benchmark, self.policy))
                self.q.put(state)
```

compiler_gym/leaderboard/llvm_instcount.py
```diff
@@ -253,33 +264,50 @@ def main(argv):
...
                 print(
                     f"=== Evaluating policy on "
                     f"{humanize.intcomma(total_count)} "
                     f"{FLAGS.test_dataset} benchmarks ==="
                     "\n\n"  # Blank lines will be filled below
                 )
-                while worker.is_alive():
-                    done_count = len(worker.states)
+
+                while len(states) < len(benchmarks):
```
ChrisCummins, line 297: There might be a bug here. If you check the line `if FLAGS.resume and Path(FLAGS.leaderboard_results).is_file():` above, the `len(benchamrks)` might be less than the `len(states)` when resuming from an incomplete run (not entirely sure, would need to test :) )

ChrisCummins (commented):
This is great! Thanks @yuandong-tian. Couple of minor suggests:

(1) I'm not sure that we should try and do any error handling through a retry loop. It can hide something that is broken. When I tried out your PR on the random search script, it just entered an infinite loop because of a `UnparsedFlagAccessError` exception being raised and silently re-queued. Try it out using:

```
cd leaderboard/llvm_instcount/random_search
python random_search.py --n=1 --max_benchmarks=3 --search_time=5
```

(2) The base case where `num_processes=1` should not use any multiprocessing. Multiprocessing can cause some unexpected behaviors in Python. E.g. the failure above is caused by the absl library not having support for multiprocessing (a [known problem](https://github.com/abseil/abseil-py/issues/107)). I left a couple of inline comments where this is affected

ChrisCummins:
Closing as stale. Feel free to re-open :)

PR #263, opened 2021-05-12
[runtime] Add an in-memory cache for Benchmark protos.

ChrisCummins:
This will be used by the CompilationSession runtime to keep track of
the Benchmark protobufs that have been sent by the user to the
service, so that `CompilationSession::init()` can be passed a benchmark
proto.

This is a generalization of the BenchmarkFactory class that is used by
the LLVM service to keep a bunch of llvm::Modules loaded in
memory. The same class is implemented twice in C++ and Python using
the same semantics and with the same tests.

The cache has a target maximum size based on the number of bytes of
its elements. When this size is reached, benchamrks are evicted using
a random policy. The idea behind random cache eviction is that this
cache will be large enough by default to store a good number of
benchmarks, so exceeding the max cache size implies a training loop in
which random programs are selected from a very large pool, rather than
smaller pool where an LRU policy would be better.

Issue #254.


compiler_gym/service/runtime/BenchmarkCache.h
```diff
@@ -0,0 +1,54 @@
...
+
+// An in-memory cache of Benchmark protocol buffers.
+//
+// This object caches Benchmark messages by URI. Once the cache reaches a
+// predetermined size, benchmarks are evicted randomly until the capacity is
+// reduced to 50%.
+class BenchmarkCache {
+ public:
+  BenchmarkCache(std::optional<std::mt19937_64> rand = std::nullopt,
+                 size_t maxSizeInBytes = kEvictionSizeInBytes);
```
hughleat: Totally not a thing, but should the order of these be reversed? Aren't they more likely to give a max size than a rand?

compiler_gym/service/runtime/BenchmarkCache.h
```diff
@@ -0,0 +1,54 @@
...
+  void add(const Benchmark&& benchmark);
+
+  inline size_t size() const { return benchmarks_.size(); };
+  inline size_t sizeInBytes() const { return sizeInBytes_; };
+  inline size_t maxSizeInBytes() const { return maxSizeInBytes_; };
+
+  void setMaxSizeInBytes(size_t maxSizeInBytes);
+
+  // Evict benchmarks randomly to reduce the capacity below 50%.
+  void prune(std::optional<size_t> targetSize = std::nullopt);
```
hughleat: Why 50%? 
Why not `evictToCapacity` or something where you can easily say - hey, I need 10kb.

compiler_gym/service/runtime/BenchmarkCache.cc
```diff
@@ -0,0 +1,83 @@
...
+using grpc::StatusCode;
+
+namespace compiler_gym::runtime {
+
+BenchmarkCache::BenchmarkCache(std::optional<std::mt19937_64> rand, size_t maxSizeInBytes)
+    : rand_(rand.has_value() ? *rand : std::mt19937_64(std::random_device()())),
+      maxSizeInBytes_(maxSizeInBytes),
+      sizeInBytes_(0){};
+
+const Benchmark* BenchmarkCache::get(const std::string& uri) const {
```
hughleat, line 19: Consider (but not for long, you know what I'm like):
```c++
class BenchmarkCache {
public:
  BenchmarkCache(std::function<Benchmark(string url)> benchmarkCreatorThingie, ...)...
  ...
  Benchmark& operator[] (string url) {
    if (url in cache) return cache[url];
    auto b = benchmarkCreatorThingie(url); // May throw std::out_of_range like std::vector.at
    if (doesn't have capacity) {
      evictUntilCapacity(b.size());
    }
    cache[url] = b;
    return b;
  }
}```

File: compiler_gym/service/runtime/BenchmarkCache.cc [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+  benchmarks_.insert({benchmark.uri(), std::move(benchmark)});
+  sizeInBytes_ += size;
+}
+
+void BenchmarkCache::prune(std::optional<size_t> targetSize) {
+  int evicted = 0;
+  targetSize = targetSize.has_value() ? targetSize : maxSizeInBytes() / 2;
+
+  while (size() && sizeInBytes() > targetSize) {
+    // Select a benchmark randomly.
```
hughleat, line 61: Ignorable: I always thought the advantages of random over lru were that random doesn't need to keep the recently used list. If the data items are small that makes sense. Otherwise, isn't lru pretty much always better?

hughleat (approved):
LGTM

compiler_gym/service/runtime/BenchmarkCache.h
ChrisCummins: Sure, done :)

compiler_gym/service/runtime/BenchmarkCache.h
ChrisCummins: Good idea, done

compiler_gym/service/runtime/BenchmarkCache.cc
ChrisCummins: I figured that there are two use common cases for this cache: (1) in a tight loop over a couple hundred benchmarks that will all fit in cache (2) over a massive set of training programs where the chance of a cache hit is negligible.

Given that, it didn't seem to me like LRU would be much of an advantage. I also found this interesting: "A random eviction policy degrades gracefully as the loop gets too big." from https://danluu.com/2choices-eviction/

Disclaimer: I know nothing about cache eviction policies and have utterly no idea what I'm talking about :)

compiler_gym/service/runtime/BenchmarkCache.cc
ChrisCummins: I don't know if that makes sense for this use case because there is no `std::function<Benchmark(string url)> benchmarkCreatorThingie` callback that would make sense. If the benchmark can't be found, we return an error to the frontend. Pseudo-code:

```py
class CompilerGymRuntime:
    def session_rpc_endpoint(self, benchmark_uri):
        if benchmark_uri not in self.benchmark_cache:
            return ErrorCode.NOT_FOUND
        benchmark = self.benchmark_cache[benchmark_uri]
        ...
```

PR #266, opened 2021-05-12
[env] Invert observation space indices.

ChrisCummins:
Reverse the order of request observation spaces so that we can pop
from the tail of the list rather than the front.


ChrisCummins:
Superseded by #271.

PR #267, opened 2021-05-12
[llvm] Add an IrSha1 observation space.

ChrisCummins:
This adds a new `IrSha1` observation space that is a 40-digit SHA1 checksum of the current module state.

```py
>>> env = gym.make("llvm-v0")
>>> env.reset()
>>> env.observation["IrSha1"]
'859b5ee6ee673571b12e8400ca09a0958a7574ec'
>>> env.step(env.action_space.sample())
(None, None, False, {'action_had_no_effect': False, 'new_action_space': False})
>>> env.observation["IrSha1"]
'937fea8de5fb0fa24ea4307cdc65df8fef188b9d'
```

This checksum is computed on the service side using LLVM's C++ implementation. It is much faster than serializing the IR and computing a checksum using python's `hashlib` or similar.


hughleat (approved):
LGTM

PR #270, opened 2021-05-13
[service] Add a reusable CompilationService runtime.

ChrisCummins:
This adds a new CompilerGymService class in both C++ and Python that
takes a concrete CompilationSession subclass and provides all of the
runtime logic to start and manage and RPC server that responds to
requests and dispatches to CompilationSession instances.

Issue #254.


ChrisCummins:
The reason there are no tests for the new code is because the next PR will port the existing compiler gym services to this new runtime, allowing the existing tests to cover this.

PR #272, opened 2021-05-16
Add a module of wrappers for CompilerEnv environments.

ChrisCummins:
This adds a series of helper wrappers under a new
`compiler_gym.warppers` module that makes it easier to modify the
behavior of CompilerEnv environments without having to change the
underlying implementation.

New classes are:

* `CompilerEnvWrapper` an extension to `gym.Wrapper` to support the
  custom compiler_gym API calls.
* `ActionWrapper` compatibility with `gym.ActionWrapper`.
* `CommandlineWithTerminalAction` adds an "end of episode" action to a
  commandline space.
* `ConstrainedCommandline` allows a subset of commandline flags to be
  selected for use.
* `CycleOverBenchmarks` loop over a list of benchamrks on `reset()`.
* `IterateOverBenchmarks` same as above but the iterator is
  exhaustible.
* `RandomOrderBenchmarks` same as above but the order is random and
  non-terminating.

### DIFF
--- a/compiler_gym/BUILD
+++ b/compiler_gym/BUILD
@@ -17,6 +17,7 @@ py_library(
         "//compiler_gym/envs",
         "//compiler_gym/leaderboard",
         "//compiler_gym/util",
+        "//compiler_gym/wrappers",
     ],
 )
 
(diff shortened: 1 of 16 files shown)


JD-ETH (approved):
Super useful, exactly what I have been mimicing :D This gives a modular way to experiment with different action/env configurations. 

Can we add the time limit wrapper as well? Fixed horizon is a very standard setup.

ChrisCummins:
> Can we add the time limit wrapper as well? Fixed horizon is a very standard setup.

Good idea, done.

PR #279, opened 2021-05-18
[examples] Update Actor Critic to use the new wrappers classes.

ChrisCummins:
There are a new set of wrappers classes that negate the need for a `CustomEnv` in the actor critic example by allowing each of the required transforms to be described by their own wrapper class.

### DIFF
--- a/examples/actor_critic.py
+++ b/examples/actor_critic.py
@@ -35,6 +35,7 @@ from absl import app, flags
 from torch.distributions import Categorical
 
 import compiler_gym  # noqa Register environments.
+from compiler_gym.wrappers import ConstrainedCommandline, TimeLimit
 
 flags.DEFINE_list(
     "flags",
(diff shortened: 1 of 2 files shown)


examples/actor_critic.py
```diff
@@ -401,11 +383,6 @@ def main(argv):
         print(f"   Avg performance: {statistics.mean(performances):.2f}")
         print(f" Worst performance: {min(performances):.2f}")
 
-    finally:
-        # The program will not terminate until the environment is
-        # closed.
-        env.close()
```
broune, line 407: Oh nice, I take it this behavior of env is no longer the case? (how?)

broune (approved):
Nice.

examples/actor_critic.py
ChrisCummins: It is still the case, but I found that gym environments have support for "with statement" usage which automatically closes the environment, so

```py
with gym.make("...") as env:
    do_stuff(env)
```

is equivalent to:

```py
try:
    env = gym.make("...")
    do_stuff(env)
finally:
    env.close()
```

PR #285, opened 2021-05-28
[examples] Port the explore script to the new wrappers classes.

ChrisCummins:
Use the new`ConstrainedCommandline` environment wrapper rather than defining a custom environment. This is the same refactor as in fa6877dcbc0b94b4208a61c0962a3f98f1f733a3 applied to the exploration script.


broune (commented):
Looks good to me.

(I'm a little bit skeptical of ".n" as meaning "count" or "number of", but that pre-existing concept isn't being introduced here, so that's orthogonal.)

PR #313, opened 2021-07-09
Add a mechanism for sending arbitrary data payloads to the backend

### LINKED ISSUE #312: Add a mechanism for sending arbitrary data payloads to the backend
## 🚀 Feature

This feature request is to add a mechanism to the CompilerGym client/service architecture so that the client can send arbitrary data messages to the service, which can then generate a response. This can be used by particular client/service implementations to provide a "back door" for talking to the service in a way that doesn't conform to the normal RL environment communication pattern.

## Motivation

The [CompilerGymService](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/service/proto/compiler_gym_service.proto#L20-L47) defines the set of messages that are needed to implement the `CompilerEnv` functionality and expose a compiler optimization problem for RL. However, it may sometimes be useful to be able to send other types of messages to the backend. Examples include:

* Configuring the maximum size of a service's in-memory cache.
* Telling the backend to disable or enable some strict runtime checks that may be useful for debugging.
* Configuring the compiler with some "meta action" that is not exposed by the compiler's action space. For example, configuring the compiler to run in debugging mode.
* Configuring the number of threads that the backend uses for handling requests.

## Pitch

1. Extend the CompilerGymService to support sending <key, value> string messages to specific sessions:

```proto
service CompilerGymService {
  ....
  // Transmit <key, value> parameters to a session.
  // Each parameter generates a string response. It us up to the
  // client/service to agree on the schema for these parameter. An
  // unknown key/value returns StatusCode::INVALID_ARGUMENT.
  rpc SendSessionParameter(SendSessionParameterRequest) returns (SendSessionParameterReply);
  ...
}

message SessionParameter {
   string key = 1;
   string value = 2;
}

message SendSessionParameterRequest {
  // The ID of the session.
  int64 session_id = 1;
  // A list of <key, value> parameters to pass to the session.
  repeated SessionParameter parameter = 2;
}

message SendSessionParameterReply {
  // A list of response messages, one per parameter request.
  repeated string reply = 1;
}
```

2. Extend the CompilationSession classes to add a session handler callback. E.g. for Python:

```py
class CompilationSession:
    def handle_session_parameter(self, key: str, value: str) -> Optional[str]:
        """Handle a session parameter send by the frontend. If the key is unknown, return None."""
        pass
```

Specific compilers can then implement this method to handle custom session parameters.

3. Extend the CompilerGymService classes to add support for this new RPC. E.g. for Python:

```py
class CompilerGymService(CompilerGymServiceServicerStub):
    ...
    def SendSessionParameter(self, request: SendSessionParameterRequest, context) -> SendSessionParameterReply:
        reply = SendSessionParameterReply()
        # Session the requested session.
        session = self.sessions[request.session_id]
        for param in request.parameter:
            # Dispatch parameter to the sesison.
            resp = session.handle_session_parameter(param.key, param.value)
            if resp is None:
                # We can add support for 'builtin' params here, e.g. maybe
                # some configuration option for the runtime. Else we report
                # the unknown param to the user.
                if resp.key == "set_some_global_thingy":
                    self.some_global_thing = param.value
                    resp = "my response"
                else:
                    raise ValueError(f"Unknown key: {param.key}")
            reply.reply.extend([resp])
        return reply
```

4. Expose this new RPC through the CompilerEnv class:

```py
class CompilerEnv(gym.Env):
    ...

    def send_session_param(self, key: str, value: str) -> str:
        """Send a <key, value> param and return response."""
        return next(self.send_session_params({key: value}).values())

    def send_session_params(self, params: Dict[str, str]) -> Dict[str, str]:
        """Send the given <key, value> params and receive <key, response> reply."""
        ...
```

A few notes on this proposal:
* I chose the name "SessionParameter" to indicate intended use. The whole point of this API is to provide a back door to send ad-hoc messages so it doesn't have to be a "parameter". Could consider another name?
* Parameters have to be sent to a specific session so that it is easy to handle them using the existing CompilationSession API (see step 2 above). For global settings (like enabling some debug mode) it might not make sense to target a specific session but I don't see an easy alternative. A caveat of this is that you can't send one of these messages unless there is an active session to send it to.
* Why send `<key, value>` pairs? Why `strings`? Why return a `string`? Happy to consider alternatives. It "feels" to me like a good balance between under-specifying the types to make it more flexible (e.g. just pass binary arrays around and have to write your own encoder/decoder for everything) and over-specifying the types in a way that makes it too inflexible / difficult to use. E.g.

```py
("benchmarks.set_max_cache_size_mb", "50")  ->  ""
("benchmarks.get_max_cache_size", "") -> "50"
("llvm.set_debug_mode", "1") -> ""
("llvm.get_unroll_run_count", "") -> "1234"
("gcc.set_real_value_param", "0.1231") -> ""
```

## Alternatives

There are two existing "back doors" for sending data to the backend: command line arguments, and environment variables. We could consider adding a mechanism for passing information that way, e.g.

```
CACHE_SIZE=12 ./my_compiler_service --port=123 --working_dir=/tmp --max_threads=4
```

but these of course allow only 1-way information, can only be set at service startup, and are not specific to a session.

ChrisCummins:
This patch adds a mechanism to the CompilerGym client/service                                     
architecture so that the client can send arbitrary data messages to                               
the service, which can then generate a response. This can be used by                              
particular client/service implementations to provide a "back door" for                            
talking to the service in a way that doesn't conform to the normal RL                             
environment communication pattern.                                                                
                                                                                                  
Fixes #312.


ChrisCummins:
Updated based on @hughleat's feedback:

1. `send_params()` always returns a list, rather than returning a singular value if called with a single parameter.
2. Added a new `send_param()` API to be clear about singular param calls.

Cheers,
Chris

PR #335, opened 2021-07-21
[www] Fix reward space calculation.

ChrisCummins:
This fixes a bug where the reward space was hard coded so that the user-requested reward space was not used.


sahirgomez1 (approved):
Thanks! LGTM

PR #336, opened 2021-07-21
[www] Truncate very-large IR strings.

ChrisCummins:
Limit the size of the IR string to ~256kB of data.


sahirgomez1 (approved):
LGTM Thank you.

PR #349, opened 2021-08-03
[www] UI components to handle url params and command line input

### LINKED ISSUE #343: Get steps from URL parameters.
Land on an specific path (command line) through URL parameters.

sahirgomez1:
### New UI components

This PR.

- Fixes #343
- Adds ability to share link with params to duplicate an environment and run specific set of actions.
- Installs react-router to handle routing and url params.
- Adds a fallback for broken links.
- Refactors frontend code in Controls Container.
- Fixes responsiveness in small screes.
- Command line input to generate and run actions.
- Refactors reward history UI as always visible.
- Allows clicking on any node to replay episode.
- Adds historical values table to observations panel.
- Adds description to observations.

https://user-images.githubusercontent.com/84393628/128580665-0cc732d5-ae16-4d8b-a684-a4a1080450e1.mov

https://user-images.githubusercontent.com/84393628/128708178-412447d8-5fb3-44b5-8ca6-9a105062b2ac.mov

https://user-images.githubusercontent.com/84393628/128921861-25bcdcd2-6dcd-4fa0-ab14-dcaf7b9cd2f7.mov


ChrisCummins:
Looking great @sahirgomez1 ! One thing - would it be possible to use the window hash to set the parameters to load from, rather than the URL proper? E.g.

```
localhost:3000/#benchmark=foo&reward=bar&actions=100
```

rather than:

```
localhost:3000/benchmark%foo/reward%bar/actions=100
```

that way, the URL can be updated as the user interacts with the tool without having to reload the page. That would enable a user to just copy the current URL to grab the state, rather than needing to click a share button to get a URL.

Cheers,
Chris

(BTW this isn't merge blocking at all :-) )

sahirgomez1:
> Looking great @sahirgomez1 ! One thing - would it be possible to use the window hash to set the parameters to load from, rather than the URL proper? E.g.
> 
> ```
> localhost:3000/#benchmark=foo&reward=bar&actions=100
> ```
> 
> rather than:
> 
> ```
> localhost:3000/benchmark%foo/reward%bar/actions=100
> ```
> 
> that way, the URL can be updated as the user interacts with the tool without having to reload the page. That would enable a user to just copy the current URL to grab the state, rather than needing to click a share button to get a URL.
> 
> Cheers,
> Chris
> 
> (BTW this isn't merge blocking at all :-) )

Sure I can make those changes and merge when it's ready.

www/frontends/compiler_gym/src/components/Modals/LargeModal.js
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React, { useState, useEffect, useRef } from "react";
+import { Modal, Button, FormControl, Row, Col } from "react-bootstrap";
+
+const LargeModal = ({ showModal, handleModal, title, getShareLink }) => {
```
s0mya, line 10: Is there a better name than LargeModal?  Is this an expanded Chart?

s0mya (approved):
@sahirgomez1 this is looking great! Thank you!

The only suggestion I would make is prefer using Typescript instead of plan js. It's the direction we're moving towards as a team.

sahirgomez1:
> @sahirgomez1 this is looking great! Thank you!
> 
> The only suggestion I would make is prefer using Typescript instead of plan js. It's the direction we're moving towards as a team.

Yep, research interfaces team will focus on Typescript.

www/frontends/compiler_gym/src/components/Modals/LargeModal.js
sahirgomez1: It's just a name for a reusable modal component, but we can definitely rename it, in fact we are not reusing it.

ChrisCummins:
LGTM! These are great improvements. Please merge when you're happy :-)

Cheers,
Chris

PR #352, opened 2021-08-10
Integrate www into packaging scripts

ChrisCummins:
* Add Flask to the python dependencies that are installed by `make init`.
* Add a new script `www/run.py` which launches the Flask API and frontend service in a single command.
* Add a `make www` target which runs the above script.


sahirgomez1 (approved):
LGTM. Yes, flask_cors is important, specially for those wanting to run locally fast. Thanks Chris.

PR #358, opened 2021-08-18
[www] Backend API hardening

ChrisCummins:
1. Make accesses to the global `sessions` table thread-safe so that concurrent users won't cause conflicts.
1. Fix a resource leak in the `/describe` API endpoint.
1. Add a background watchdog thread that garbage collects sessions that haven't been use for 20 minutes.
1. Add @hughleat's code to serve the static frontend assets over Flask.
1. Simplify the build instructions by making `make www` build the static assets and then launch the Flask app, rather than using `npm run` and flask simultaneously.
1. Rename `demo_api.py` to `www.py` since it now serves both backend API and frontend assets.


s0mya (approved):
LGTM, Thanks!

PR #314, opened 2021-07-12
Gcc Port for CompilerGym

hughleat:
This port adds an environment to Compiler Gym to enable GCC to be used instead of LLVM.

PR #298, opened 2021-06-22
[loop_tool] Add integration and tests

bwasti:
Update: it's working :)

```
python -m compiler_gym.bin.random_search --env=looptool-v0 --benchmark=loop_tool-v0/8388608 --reward=flops --nproc=1 --runtime=60
```

This PR scaffolds integration of the ~(nearly open sourced)~ loop_tool.  ~This doesn't include the full integration (work in progress).~

I wanted to put this up to get feedback on the proposed integration points.  A couple questions

1. should this live somewhere else? envs/loop_tool seems reasonable, but the changes required to the build and requirements.txt seem somewhat invasive
2. if this should live in envs/loop_tool, should I try to refactor things to make loop_tool optional? (basically just remove from requirements.txt and add a fallback error message asking folks to `pip install loop_tool`)
3. Is CompilerSession really an env? I think I might be getting semantics wrong here, but I'm not sure the exact parallels to llvm's integration

PR #359, opened 2021-08-18
[docs] Add a new "Environments" section to docs.

ChrisCummins:
Adds a new top-level section to the documentation. This is to make room for new documentation pages for the GCC environment reference (#314) and loop_tool (#298).

image


ChrisCummins:
FYI @hughleat @bwasti

PR #361, opened 2021-08-20
[www] Package the Flask app as a Docker container

ChrisCummins:
This adds a new `make www-image` target that builds and runs the CompilerGym Explorer web frontend in a Docker image. The image is self-contained and can be deployed elsewhere.


sahirgomez1 (approved):
Thanks Chris, I will need a couple of minutes from you to set this up locally. Happy to schedule a meeting.

Issue #52, opened 2021-02-04
Add support for discrete and continuous action spaces


ChrisCummins:
## 🚀 Feature

Add the ability to describe discrete and continuous action spaces, with optional bounds.

## Motivation

Currently, the [ActionSpace](https://github.com/facebookresearch/CompilerGym/blob/93ae6ce44b0a8d5a1c46de4dbee5fd591fb699c7/compiler_gym/service/proto/compiler_gym_service.proto#L94-L103) specification supports only categorical spaces. This is fine for some problems, such as [LLVM phase ordering](https://facebookresearch.github.io/CompilerGym/llvm/index.html#action-space), but is not sufficiently expressive for many other optimizations such as unroll factors, tiling sizes, inlining thresholds, etc.

ChrisCummins:
cc @hughleat  @bwasti we may want to revisit this to better support GCC / loop_tool.

ChrisCummins:
Done.

PR #369, opened 2021-09-02
Extend action spaces to be composable and continuous.

ChrisCummins:
This pull request changes the way in which action spaces are exposed by compiler services so that a more expressive set of action spaces can be described, such as tuple spaces and continuous spaces.

## Background

A `CompilerEnv` queries the set of action spaces that a particular compiler supports by calling a `GetSpaces()` RPC endpoint. This returns a list of [ActionSpace](https://github.com/facebookresearch/CompilerGym/blob/514711130b3c2b8b873dfa05fbd979d7168ebb34/compiler_gym/service/proto/compiler_gym_service.proto#L132-L137) protocol buffers. Currently the action space schema only supports a list of string action names:

https://github.com/facebookresearch/CompilerGym/blob/514711130b3c2b8b873dfa05fbd979d7168ebb34/compiler_gym/service/proto/compiler_gym_service.proto#L132-L137

This is fine for discrete spaces like command-line flags:

```py
ActionSpace(
    name="command_line_flags",
    action=[
        "-adce",
        "-loop-unroll",
        "-inline",
    ],
)
```

but is not well suited for spaces where the choices are not discrete, or spaces where an action may comprise multiple choices. For example, suppose we would like to expose a compiler's [loop unroll](https://en.wikipedia.org/wiki/Loop_unrolling) factor as an action space. We would need to map the integer choices to a discrete space like so:

```py
ActionSpace(
    name="loop_unroll_factor",
    action=[
        "1",
        "2",
        "4",
        "8",
        ...
    ],
)
```

or suppose an action space consisted of two independent boolean choices (e.g. "inline" and "unroll"). To fit the current action space design, the cartesian product of these two choices must be exposed:

```py
ActionSpace(
    name="two_choices",
    action=[
        "inline=false,unroll=false",
        "inline=false,unroll=true",
        "inline=true,unroll=false",
        "inline=true,unroll=true",
    ],
)
```

as more choices are added, the size of the space will blow up exponentially.

## Changes

To address these shortcomings outlined above this pull request expands the `ActionSpace` schema to support:

1. composable actions: actions which comprise multiple independent choices
2. numeric actions: (optionally) bounded integer and real-valued choices

### Composable Actions

The new action space schema decomposes an action into a list of one or more choices:

```proto
message Action {
  // A list of choices that corresponds to the ActionSpace's choice list.
  repeated Choice choice = 1;
}
```

where the choices that are available for an action are defined a priori by the `ActionSpace` message. For example, suppose we have an action space that comprises two independent choices:

```py
ActionSpace(
    name="opt_and_debug",
    choice=[
        ChoiceSpace(
            name="opt",
            named_discrete_space=NamedDiscreteSpace(
                value=[
                    "-O0",
                    "-O1",
                    "-O2",
                    "-O3",
                ]
            )
        ),
        ChoiceSpace(
            name="debug",
            named_discrete_space=NamedDiscreteSpace(
                value=[
                    "-g0",
                    "-g",
                ]
            )
        ),
    ]
)
```

this will be exposed by CompilerGym as a `gym.Tuple` space that can be used to specify both flags simultaneously in a single `step()`:

```py
env = gym.make(...)
env.reset()
env.step((0, 1))  # -O0, -g
env.step((2, 0))  # -O2, -g0
```

An additional flag `ActionSpace.named_choices` can also be used to instead expose the choices as a `gym.Dict` space, allowing each choice to be specified by name:

```py
env = gym.make(...)
env.reset()
env.step({"opt": 0, "debug": 1})  # -O0, -g
env.step({"opt": 2, "debug": 0})  # -O2, -g0
```

### Numeric actions

In addition to the `NamedDiscreteSpace` that is well suited for command line flags, two new scalar types have been added to support optionally bounded integer or double choices.

For example, suppose you would like to expose a loop unrolling threshold in the range [0, 1]:

```py
ActionSpace(
    name="loop_unroll_threshold",
    choice=[
        ChoiceSpace(
            name="threshold",
            double_range=ScalarRange(
                min=ScalarLimit(value=0),
                max=ScalarLimit(value=1),
            )
        )
    ]
)
```

This can then be used as with other action types:

```py
env = gym.make(...)
env.reset()
env.step(0.3)
env.step(0.9)
```

These new action types can be composed with others:

```py
ActionSpace(
    name="multi_space",
    choice=[
        ChoiceSpace(
            name="loop_unroll_factor",
            int64_range=ScalarRange(
                min=ScalarLimit(value=1),
                max=None,  # no upper bound
            )
        ),
        ChoiceSpace(
            name="loop_unroll_threshold",
            double_range=ScalarRange(
                min=ScalarLimit(value=0),
                max=ScalarLimit(value=1),
            )
        ),
        ChoiceSpace(
            name="flag",
            named_discrete_space=NamedDiscreteSpace(
                value=[
                    "-foo",
                    "-bar",
                    "-inline",
                ]
            )
        )
    ]
)
```

## Performance Impact

These changes add a number of new complications to `StepRequest` encoding and decoding. It replaces a list of integers with a list of nested message types, each with a union-type field. On the client side, this increases the number of protocol buffer heap allocations by `n * m` per step (`n` = num actions, `m` = num choices per action). On the server side, the `apply_action()` code has to be a little more complex with additional error handling for the corner cases of malformed messages.

The size of the performance overhead depends on the runtime of the environment. Measuring the average step times of environments in micro-benchmarks reveals:

*  **1% regression** in step time for a "slow" LLVM environment where average step time is **66.182ms**.
* **5% regression** in step time for a "fast" LLVM environment where average step time is **0.310ms**.
* **17% regression** for the worst case C++ example service where average step time is only **0.126ms** as no actual compilation happens and the step time is dominated by service overheads.

Full measurements:

```
----------------------------------------------------------- benchmark 'test_step[dummy-cc]': 2 tests ----------------------------------------------------------
Name (time in us)                           Min              Median                 Max                Mean            StdDev            OPS (Kops/s)          
---------------------------------------------------------------------------------------------------------------------------------------------------------------
test_step[dummy-cc] (development-)     109.7329 (1.0)      127.0049 (1.0)      131.2449 (1.0)      126.2900 (1.0)      3.8702 (1.0)            7.9183 (1.0)    
test_step[dummy-cc] (pr-369)           138.2914 (1.26)     145.9978 (1.15)     157.4319 (1.20)     148.0249 (1.17)     4.7854 (1.24)           6.7556 (0.85)   
---------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------- benchmark 'test_step[dummy-py]': 2 tests -----------------------------------------------------------
Name (time in us)                           Min              Median                 Max                Mean             StdDev            OPS (Kops/s)          
----------------------------------------------------------------------------------------------------------------------------------------------------------------
test_step[dummy-py] (development-)     440.8386 (1.0)      562.9815 (1.0)      592.2488 (1.0)      550.9693 (1.0)      35.1322 (1.0)            1.8150 (1.0)    
test_step[dummy-py] (pr-369)           453.1838 (1.03)     626.0769 (1.11)     667.4190 (1.13)     607.5601 (1.10)     47.8933 (1.36)           1.6459 (0.91)   
----------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------- benchmark 'test_step[llvm;fast-benchmark;fast-action]': 2 tests -----------------------------------------------------------
Name (time in us)                                                  Min              Median                 Max                Mean             StdDev            OPS (Kops/s)          
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_step[llvm;fast-benchmark;fast-action] (development-)     291.6037 (1.0)      304.8116 (1.0)      551.3841 (1.63)     309.6838 (1.0)      30.9735 (4.72)           3.2291 (1.0)    
test_step[llvm;fast-benchmark;fast-action] (pr-369)           306.1753 (1.05)     325.1733 (1.07)     337.5231 (1.0)      325.1713 (1.05)      6.5576 (1.0)            3.0753 (0.95)   
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

----------------------------------------------------------- benchmark 'test_step[llvm;fast-benchmark;slow-action]': 2 tests -----------------------------------------------------------
Name (time in us)                                                  Min              Median                 Max                Mean             StdDev            OPS (Kops/s)          
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_step[llvm;fast-benchmark;slow-action] (development-)     753.4978 (1.0)      869.4267 (1.0)      905.9991 (1.0)      873.5460 (1.0)      27.2897 (1.18)           1.1448 (1.0)    
test_step[llvm;fast-benchmark;slow-action] (pr-369)           842.4660 (1.12)     899.2449 (1.03)     942.2123 (1.04)     903.8721 (1.03)     23.1419 (1.0)            1.1064 (0.97)   
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

---------------------------------------------------- benchmark 'test_step[llvm;slow-benchmark;fast-action]': 2 tests -----------------------------------------------------
Name (time in ms)                                                Min            Median               Max              Mean            StdDev                 OPS          
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_step[llvm;slow-benchmark;fast-action] (development-)     3.8598 (1.19)     4.4526 (1.00)     4.5908 (1.02)     4.4591 (1.01)     0.0660 (1.0)      224.2628 (0.99)   
test_step[llvm;slow-benchmark;fast-action] (pr-369)           3.2326 (1.0)      4.4352 (1.0)      4.5220 (1.0)      4.3992 (1.0)      0.1585 (2.40)     227.3128 (1.0)    
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------

------------------------------------------------------ benchmark 'test_step[llvm;slow-benchmark;slow-action]': 2 tests ------------------------------------------------------
Name (time in ms)                                                 Min             Median                Max               Mean            StdDev                OPS          
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_step[llvm;slow-benchmark;slow-action] (development-)     63.5560 (1.0)      65.8887 (1.00)     95.5935 (1.04)     66.1820 (1.0)      3.4628 (1.0)      15.1098 (1.0)    
test_step[llvm;slow-benchmark;slow-action] (pr-369)           63.7788 (1.00)     65.7038 (1.0)      91.9016 (1.0)      67.0707 (1.01)     4.8654 (1.41)     14.9096 (0.99)   
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Negligible changes to the other operation times.

## Pull request status

- [x] Initial design of new action schema.
- [x] Port the example environments to use it.
- [x] Port the LLVM environments to use it.
- [x] Compare microbenchmark step time before/after.
- [x] Write tests for the new python action space converters.
- [x] Update the sphinx documentation.

Issue #52.


compiler_gym/envs/llvm/service/LlvmSession.cc
```diff
@@ -129,7 +129,10 @@ Status LlvmSession::applyAction(const Action& action, bool& endOfEpisode,
   switch (actionSpace()) {
     case LlvmActionSpace::PASSES_ALL:
       LlvmAction actionEnum;
-      RETURN_IF_ERROR(util::intToEnum(action.action(), &actionEnum));
+      if (action.choice_size() != 1) {
```
mostafaelhoushi, line 125: should the condition be `< 1`? 
Otherwise the error message `"Missing choice"` is misleading

compiler_gym/spaces/reward.py
```diff
@@ -38,6 +38,8 @@ class Reward(Scalar):
 
     def __init__(
         self,
+        # TODO(cummins): Rename `id` to `name` for consistency with the other
```
mostafaelhoushi: Do you want to implement this TODO in this PR or in a future PR?

compiler_gym/envs/llvm/service/LlvmSession.cc
ChrisCummins: Thanks, I've improved the error message

compiler_gym/spaces/reward.py
ChrisCummins: Good question. I've filed an issue and updated the comment to reference it. #381.

mostafaelhoushi:
I am wondering if you can add compiler test cases, rather than protobuf test cases. 

I mean an actual test case with the user specifying the range of loop unroll factors and/or thresholds and runing `env.step()`. Basically the same examples you have provided in the description of this PR, can they be test cases added to the repo?

compiler_gym/service/proto/compiler_gym_service.proto
```diff
@@ -174,6 +174,9 @@ message NamedDiscreteSpace {
   // A list of names for every value in the space. The order in which these
   // values are returned is used.
   repeated string value = 1;
+  // If true, the set of space can be interpreted as a list of command line
+  // flags.
+  bool is_commandline = 2;
```
hughleat, line 179: Why is this needed?

compiler_gym/service/proto/py_converters.py
```diff
@@ -0,0 +1,125 @@
...
+            if choice.int64_range.min.value:
+                return PyChoiceSpace(
+                    space=Scalar(
+                        min=choice.int64_range.min.value,
+                        max=choice.int64_range.max.value,
+                        dtype=int,
+                    ),
+                    make_choice=lambda a: Choice(int64_value=a),
+                )
+            else:
```
hughleat: Why is having a min value make it a range type, and not having a min value make it named range?

compiler_gym/envs/llvm/llvm_env.py
```diff
@@ -165,32 +165,59 @@ def __init__(
...
-                "l1d_cache_size": Scalar(min=None, max=None, dtype=int),
-                "l1d_cache_count": Scalar(min=None, max=None, dtype=int),
-                "l2_cache_size": Scalar(min=None, max=None, dtype=int),
-                "l2_cache_count": Scalar(min=None, max=None, dtype=int),
-                "l3_cache_size": Scalar(min=None, max=None, dtype=int),
-                "l3_cache_count": Scalar(min=None, max=None, dtype=int),
-                "l4_cache_size": Scalar(min=None, max=None, dtype=int),
-                "l4_cache_count": Scalar(min=None, max=None, dtype=int),
-            }
+                "name": Sequence(name="name", size_range=(0, None), dtype=str),
```
hughleat: Consider a function like make_named_dict_space, which doesn't require writing the names twice?

compiler_gym/service/proto/compiler_gym_service.proto
```diff
@@ -350,7 +351,9 @@ message AddBenchmarkReply {}
 
 // An arbitrary string parameter value for a session.
 message SessionParameter {
+  // A string key.
```
hughleat, line 371: I love these docs! ;)

compiler_gym/envs/llvm/service/LlvmSession.cc
```diff
@@ -130,7 +130,9 @@ Status LlvmSession::applyAction(const Action& action, bool& endOfEpisode,
     case LlvmActionSpace::PASSES_ALL:
       LlvmAction actionEnum;
       if (action.choice_size() != 1) {
-        return Status(StatusCode::INVALID_ARGUMENT, "Missing choice");
+        return Status(
+            StatusCode::INVALID_ARGUMENT,
+            fmt::format("Invalid choice count. Expected 1, received {}", action.choice_size()));
```
hughleat, line 128: This is just me being curious and not a review comment.
Why use `fmt::format` instead of an `f` string?

hughleat (approved):
LGTM.

But, there is a part of me that wonders about a different arrangement. Suppose that there is a concept of a compilation strategy whose data type is described with ranges etc. For LLVM it is an unbounded list of passes plus other parameters with given ranges, etc. For GCC it is just the parameters with ranges. It is somewhat similar to the structured data description you have here, but includes at least lists as well.
Then, once the strategy space is defined, actions are automatically built on top of it. So, if there is a list, like for passes, then actions can be automatically created that mutate the list. For parameters, actions can modify the individual values. An action is also there to completely replace the current compilation strategy.
A compiler service, then would report the strategy definition, documentation and starting strategy.
The client constructs a starting strategy and makes actions for it. It remembers the current strategy.
When a user makes an action, this updates the client strategy. If that makes a change (it might not always) and observations are requested, then the strategy is sent to the service for processing.
Most services would be happy to do that statelessly. The LLVM opt service would find a nearby point and execute passes to move if possible or restart if not.

This would make actions look similar across different compilers and simplify things (maybe). It would pave the way for the LLVM opt space to include more options and such, as well as bringing in llc.

Not 100% sure if this suggestion works well with changing spaces.

compiler_gym/service/proto/py_converters.py
ChrisCummins: `Discrete` is OpenAI gym's way of describing some integer range `[0,n]`. For ranges that aren't based on zero I added a `Scalar` class that covers ranges `[x,y]`. We will use `Discrete` if possible, else fallback to `Scalar`. Will add a clarifying comment.

compiler_gym/envs/llvm/service/LlvmSession.cc
ChrisCummins: C++ has no f strings :(

compiler_gym/service/proto/compiler_gym_service.proto
ChrisCummins: Oh shush you 😃

These pointless comment strings are actually pretty useful in the generated html docs where there isn't as much context / type info.

compiler_gym/service/proto/compiler_gym_service.proto
ChrisCummins: The python frontend has a `Commandline` class which extends the `NamedDiscrete` space with a couple of extra bits of functionality, like the ability to document the flags and serialize / parse from string. This field tells the frontend to treat it as a commandline.

compiler_gym/envs/llvm/llvm_env.py
ChrisCummins: Good idea, thanks!

compiler_gym/envs/llvm/service/LlvmSession.cc
hughleat: Ahem, umm, oh, yeah. 
I can't think why you'd imagine this wasn't a thorough code review.

ChrisCummins:
I've update the GCC and loop_tool environments to use this new action space. Lets merge this now and then submit follow-up PRs to port those environments to expose a more natural action space.

Cheers,
Chris

PR #418, opened 2021-09-21
[leaderboard] Change benchmarks execution to Round Robin

### LINKED ISSUE #243: Use round-robin ordering for leaderboard experiment execution
## 🚀 Feature

Don't repeat all `--n` runs of each benchmark in order, perform 1 run of each benchmark, then proceed to the 2nd run, etc, until all `n` runs have been completed.

thecoblack:
Hi! Benchmarks were being sorted that means that we had something like `[1, 1, 2, 2, 3 ,3]`, so if we remove the sort and leave it as it is we'll have `[1, 2, 3, 1, 2, 3]`. I think this is what was expected after reading the description. Please let me know what do you think :)!

Signed-off-by: thecoblack <victoracb13@gmail.com>

Fixes #243

### DIFF
--- a/compiler_gym/leaderboard/llvm_instcount.py
+++ b/compiler_gym/leaderboard/llvm_instcount.py
@@ -246,7 +246,6 @@ def eval_llvm_instcount_policy(policy: Policy) -> None:
 
             # Repeat the searches for the requested number of iterations.
             benchmarks *= FLAGS.n
-            benchmarks = sorted(benchmarks)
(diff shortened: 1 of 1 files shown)


ChrisCummins:
You nailed it, thanks @thecoblack!

Normally I would be picky in demanding a test, but this module isn't currently well structured for testing so I think that'll be fine.

Thank you for making your first contribution to CompilerGym!

Cheers,
Chris

PR #432, opened 2021-09-27
Unrolling example service

mostafaelhoushi:
## Summary
Simple example service that:
- has different unrolling factors as actions
- creates a simple dataset with 2 .c files
- we use a `BENCH` C function within the benchmarks that prints the measured time to stdout and it gets parsed to measure the reward. 
 
`BENCH` is used in famous benchmakrs like gcc-loops and TSVC, so hopefully that is a nice way to reuse benchmarking tools already existing out there.

## TODOs
To complete the PR, I need to update the README in the unrolling example directory with details on how to run the example, how to modify it, and how to create benchmarks.

## Issues
It seems that `clang` or `opt` are ignoring the unrolling factor when we pass `-unroll-count=<NUMBER>`. The only way to force an unrolling factor is to add it as `#pragma` statement in the C code before the loop. 
So we have different options to deal with this:
1. Continue debugging and find a way to pass the unrolling factor to the command-line
2. Modify `apply_action` to insert the #pragma into the source file. I think that is what https://github.com/intel/neuro-vectorizer has done. The advantage is that it will allow users to have fine grain control on each loop in the code. The disadvantage is that it will make the example too complicated. Maybe that could be a separate project or separate PR.
3. Look for another pass whose parameters can be controlled from a command-line option

## Future
We can either enhance this example, or create new examples by:
1. Adding more loop transformations with their parameters. (Use the recently added feature of composable actions)
2. Allowing fine grain control on loops by inserting #pragma statments into the source code
3. Add observations that are related to loop transformations, such as polyhedral access matrices

### DIFF
--- a/examples/example_unrolling_service/env_tests.py
+++ b/examples/example_unrolling_service/env_tests.py
@@ -22,10 +22,10 @@ from tests.test_main import main
 
 # Given that the C++ and Python service implementations have identical
 # featuresets, we can parameterize the tests and run them against both backends.
-EXAMPLE_ENVIRONMENTS = ["unrolling-py-v0"]
+UNROLLING_ENVIRONMENTS = ["unrolling-py-v0"]
 
 
-@pytest.fixture(scope="function", params=EXAMPLE_ENVIRONMENTS)
+@pytest.fixture(scope="function", params=UNROLLING_ENVIRONMENTS)
 def env(request) -> CompilerEnv:
     """Text fixture that yields an environment."""
     with gym.make(request.param) as env:
@@ -41,7 +41,7 @@ def bin(request) -> Path:
     yield request.param
 
 
-@pytest.mark.parametrize("env_id", EXAMPLE_ENVIRONMENTS)
+@pytest.mark.parametrize("env_id", UNROLLING_ENVIRONMENTS)
 def test_debug_level(env_id: str):
     """Test that debug level is set."""
     set_debug_level(3)
--- a/examples/example_unrolling_service/env_tests.py
+++ b/examples/example_unrolling_service/env_tests.py
@@ -1,0 +1,232 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""Tests for the example CompilerGym service."""
+import logging
+import subprocess
+from pathlib import Path
+
+import gym
+import numpy as np
+import pytest
+from gym.spaces import Box
+
+import compiler_gym
+import examples.example_unrolling_service as example
+from compiler_gym.envs import CompilerEnv
+from compiler_gym.service import SessionNotFound
+from compiler_gym.spaces import NamedDiscrete, Scalar, Sequence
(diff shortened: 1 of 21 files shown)


examples/example_unrolling_service/service_py/example_service.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __init__(
+        self, working_directory: Path, action_space: ActionSpace, benchmark: Benchmark
+    ):
+        super().__init__(working_directory, action_space, benchmark)
+        logging.info("Started a compilation session for %s", benchmark.uri)
+        self._benchmark = benchmark
+        self._update_observations()
+
+    def _update_observations(self):
```
mostafaelhoushi: @ChrisCummins : I am not sure if what I am doing here is correct. I am trying to obtain the LLVM IR from the bitcode. I am worried that I am gradually re-implementing the llvm environment.

Should I:
- continue with this path?
- can I use reuse LLVM benchmark or environment to obtain the IR (and other observations as well)?
- try to do something easier where I obtain the LLVM IR from source file rather than from bitcode? The point is that I am using `llvm_benchmark.make_benchmark(...)` which creates a bitcode file and I can't get a hand of the source file.

ChrisCummins: I would pick option (3). I think for your use case, a "benchmark" is best represented as the C source file, so use `Benchmark.from_file()` rather than `llvm.make_benchmark()`. Then this method can run clang on the C code to generate LLVM-IR.

Cheers,
Chris

examples/example_unrolling_service/service_py/example_service.py
```diff
@@ -0,0 +1,162 @@
...
+        self._observation = dict()
+        # FIXME: "llvm-dis" path is not found. Should we add its path, or is there another way to obtain the IR?
```
ChrisCummins: In case you haven't seen, [this module](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/third_party/llvm/__init__.py#L85-L87) contains some useful functions for getting paths to LLVM tools, although we currently don't ship llvm-dis.

examples/example_unrolling_service/env_tests.py
```diff
@@ -0,0 +1,239 @@
...
+    env.step(env.action_space.sample())
+    assert env.in_episode
+
+
+def test_Step_out_of_range(env: CompilerEnv):
+    """Test error handling with an invalid action."""
+    env.reset()
+    with pytest.raises(ValueError) as ctx:
+        env.step(100)
+    assert str(ctx.value) == "Out-of-range"
```
mostafaelhoushi, line 150: @ChrisCummins : I am almost done with the PR. One last thing remaining is this test case. Instead of it raising `ValueError: Out-of-range` it is instead raising `FileNotFoundError: list index (3) out of range`. Not sure why

mostafaelhoushi: It is now solved after your changes. Will resolve.

mostafaelhoushi:
Thanks Chris for your changes. They look good to me. I am just updating the README file now.

PR #433, opened 2021-09-27
[ci] Add a codeql workflow for Python.

ChrisCummins:


ChrisCummins:
It looks like this workflow doesn't work without a base run on the `development` branch, so doing that now.

PR #442, opened 2021-09-29
[llvm] Add a new Bitcode observation space.

### LINKED ISSUE #426: Observation 'Bitcode' without going through a temporary file.
## 🚀 Feature

Have an observation which just returns the bytes of the bitcode, rather than the file. Going through a temporary file is a (tiny) bit annoying.

```
env.observation["Bitcode"] -> bytes
env.bitcode == env.observation["Bitcode"]
```
## Motivation

I'm passing bitcode to LLVM tools. They will accept bitcode on stdin and emit it to stdout. No need for a temporary file.

## Pitch

Don't use a file when a `bytes` will do? 

## Alternatives

Keep using a file.

ChrisCummins:
This adds a Bitcode observation space that returns the bitcode of an
LLVM module directly as a byte array, without having to go via an
intermediate file.

Fixes #426.

### DIFF
--- a/compiler_gym/envs/llvm/service/Observation.cc
+++ b/compiler_gym/envs/llvm/service/Observation.cc
@@ -59,6 +59,13 @@ Status setObservation(LlvmObservationSpace space, const fs::path& workingDirecto
       reply.set_string_value(ss.str());
       break;
     }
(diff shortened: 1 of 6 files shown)


hughleat (approved):
Perfect :)

PR #446, opened 2021-09-29
[datasets] Add a weighted implementation of random_benchmark().

### LINKED ISSUE #422: Possible error in how to generate unbiased benchmarks
## 📚 Documentation
(Possible error, possibly me doing it wrong)
https://facebookresearch.github.io/CompilerGym/compiler_gym/datasets.html

The code for `random_benchmark` says:

```python
rng = np.random.default_rng()
finite_datasets = [d for d in env.datasets if len(d) != math.inf]
dataset = rng.choice(
    finite_datasets,
    p=[len(d) for d in finite_datasets]
)
dataset.random_benchmark(random_state=rng)
```

But, I think that if you call `len` on a generator, it will fail with error:
`TypeError: 'float' object cannot be interpreted as an integer`

Suggest changing it to:
`d.size != math.inf`

ChrisCummins:
By default datasets are selected uniformly randomly in
Datasets.random_benchmark(). This means that datasets with a small
number of benchmarks will be overrepresented compared to datasets with
many benchmarks. To correct for this bias this patch adds a
weight_datasets_by_size argument which is equivalent to weighting the
dataset choice by the number of benchmarks in each dataset:

    >>> random.choices(datasets, weights=[len(p) for p in datasets])

Weighting the choice of datasets by their size means that datasets with
infinite sizes (such as random program generators) will be excluded from
sampling as their size is 0.

Fixes #422.

### DIFF
--- a/compiler_gym/datasets/datasets.py
+++ b/compiler_gym/datasets/datasets.py
@@ -3,7 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 from collections import deque
-from typing import Dict, Iterable, Optional, Set, TypeVar
+from typing import Dict, Iterable, List, Optional, Set, TypeVar
 
 import numpy as np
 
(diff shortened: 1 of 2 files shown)


hughleat (approved):
LGTM

I have a suggestion though. Instead of giving infinite datasets a weight of zero, allow that to be a parameter. This way, I can decide how often the infinite ones should contribute. 
Or the infinite ones collectively get a weight equal to the sum of the others.  Something like that.

ChrisCummins:
> I have a suggestion though. Instead of giving infinite datasets a weight of zero, allow that to be a parameter. This way, I can decide how often the infinite ones should contribute. Or the infinite ones collectively get a weight equal to the sum of the others. Something like that.

I like this! How about this:

1. I've renamed `weighted_by_datset_size` arg to just `weighted`.
2. I've added an optional `weights` arg that is a dictionary mapping dataset name to weight.
3. Weighted sampling will now use the weight provided by the weights dictionary, else default to the size of the dataset. E.g.

```py
>>> env.datasets.random_benchmark(weighted=True, weights={
    "generator://csmith-v0": 1e6,
    "benchmark://some-dataset-v0": 555,
})
```

hughleat:
I like that change, but I suggest having more than one alternative. If we **have** to give weights then it seems like the user has to do quite a bit of work. Why not additionally allow an automated weighting:

```python
>>> env.datasets.random_benchmark(weights=True, generator_ratio=0.5)
```
If the `weights` are given explicitly as a dictionary, use those.
If `weight=True` then the weights for `n` generators will be `sum(non_generator_sizes) * generator_ratio / n`
`'generator_ratio` defaults to `0`. 
`generator_ratio` must be in [0, 1).
If `weights=False or None` then choose uniformly from the datasets.

This way users can be specific or take an easy route.


An unrelated thing I was thinking is to easily allow people to get datasets that are OSS - e.g. no Anhgabench. Something like:
`dataset.is_oss` and `datasets.exclude_non_oss()`.

Why? So I can do:
```python
env.datasets.exclude_non_oss()
del env.datasets['cbench'] # to keep out for testing
b = env.datasets.random_benchmark(weights=True, generator_ratio=0.5)
train_on(b)
```

ChrisCummins:
> Why not additionally allow an automated weighting:
> 
> ```python
> >>> env.datasets.random_benchmark(weights=True, generator_ratio=0.5)
> ```
> 
> If the `weights` are given explicitly as a dictionary, use those. If `weight=True` then the weights for `n` generators will be `sum(non_generator_sizes) * generator_ratio / n` `'generator_ratio` defaults to `0`. `generator_ratio` must be in [0, 1). If `weights=False or None` then choose uniformly from the datasets.

I worry that that is too specific of a solution. If the user wants to include generators as weighted by the total number of benchmarks they can roll their own weights dictionary in a couple of lines of code:

```py
>>> n = sum(len(d) for d in env.datasets)
>>> weights = {d.name: (len(d) or 0.5 * n) for d in env.datasets}
>>> env.datasets.random_benchmark(weights=weights)
```

On the other hand, a user may want to use a fixed weight for generators, or provide per-generator weights. I'll expand the docstring to include a couple of examples of generating weights dicts.

> An unrelated thing I was thinking is to easily allow people to get datasets that are OSS - e.g. no Anhgabench. Something like: `dataset.is_oss` and `datasets.exclude_non_oss()`.
> 
> Why? So I can do:
> 
> ```python
> env.datasets.exclude_non_oss()
> del env.datasets['cbench'] # to keep out for testing
> b = env.datasets.random_benchmark(weights=True, generator_ratio=0.5)
> train_on(b)
> ```

There is a [Dataset.license](https://facebookresearch.github.io/CompilerGym/compiler_gym/datasets.html#compiler_gym.datasets.Dataset.license) attribute which contains the shorthand name of the license as reported by that dataset's LICENSE/readme file.  I wouldn't want to try and interpret those licenses as that could get very tricky quickly - anghabench is very literally "open source", and some licenses have exceptions / differences in allowed usage. In your case you could achieve your goal using:

```py
>>> [d for d in env.datasets if "Unknown" not in d.license]
[benchmark://cbench-v1,
 benchmark://blas-v0,
 benchmark://chstone-v0,
 benchmark://clgen-v0,
 benchmark://github-v0,
 benchmark://linux-v0,
 benchmark://mibench-v0,
 benchmark://npb-v0,
 benchmark://opencv-v0,
 benchmark://poj104-v1,
 benchmark://tensorflow-v0,
 generator://csmith-v0,
 generator://llvm-stress-v0]
```

or perhaps filter on a set of "approved" licenses:

```py
>>> [d for d in env.datasets if "Apache" in d.license or "BSD" in  d.license]
[benchmark://cbench-v1,
 benchmark://blas-v0,
 benchmark://mibench-v0,
 benchmark://opencv-v0,
 benchmark://poj104-v1,
 benchmark://tensorflow-v0,
 generator://csmith-v0,
 generator://llvm-stress-v0]
```

Cheers,
Chris

Issue #466, opened 2021-10-13
How to run the  Example CompilerGym Service


Angerla:
## ❓ Questions and Help

I want to run the Example CompilerGym Service, but I'm not clear how to do it. I download the source of the example, the main working directory is as follows::
```shell
|-- CompilerGym
|   |-- BUILD.bazel
|   |-- WORKSPACE
|   |-- examples
|   |   |-- example_compiler_gym_service
|   |   |   |-- BUILD
|   |   |   |-- README.md
|   |   |   |-- WORKSPACE
|   |   |   |-- __init__.py
|   |   |   |-- env_tests.py
|   |   |   |-- service_cc
|   |   |   |   |-- BUILD
|   |   |   |   `-- ExampleService.cc
|   |   |   `-- service_py
|   |   |       |-- BUILD
|   |   |       `-- example_service.py
......
```
I tried this command `bazel run -c opt //examples/example_compiler_gym_service/service -- --port=8080 --working_dir=/tmp` in the root directory "CompilerGym", but get the error:  
```shell
ERROR: Skipping '//examples/example_compiler_gym_service/service': no such package 'examples/example_compiler_gym_service/service': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /home/zc/rpc_test/CompilerGym/examples/example_compiler_gym_service/service
WARNING: Target pattern parsing failed.
ERROR: no such package 'examples/example_compiler_gym_service/service': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /home/zc/rpc_test/CompilerGym/examples/example_compiler_gym_service/service
INFO: Elapsed time: 1.316s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```
I also tried the command `bazel run -c opt //examples/example_compiler_gym_service/service_cc -- --port=8080 --working_dir=/tmp` at the root  directory and get the error:
```shell
ERROR: Skipping '//examples/example_compiler_gym_service/service_cc': no such target '//examples/example_compiler_gym_service/service_cc:service_cc': target 'service_cc' not declared in package 'examples/example_compiler_gym_service/service_cc' defined by /home/zc/rpc_test/CompilerGym/examples/example_compiler_gym_service/service_cc/BUILD
WARNING: Target pattern parsing failed.
[...]

PR #467, opened 2021-10-13
[examples] Update and improve the example service implementations and documentation

ChrisCummins:
This patch set includes:

* updates and corrects the README file for the example gym services, fixing #466.
* this adds a new `demo.py` script that includes a minimal usage of the example environments. This is intended to be used as a starting point for people who write their own services.
* adds a new `demo_without_bazel.py` script which shows how the Python service can be used independently of the bazel build system.
* fixes a bug in which `--port` flag was not used by the Python service runtime.
* fixes a minor behavior difference between the C++ / Python runtimes in that C++ required the logs directory to exist wheras the Python version would create the directory. Both runtimes now create the directory.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -18,7 +18,7 @@ On macOS the required dependencies can be installed using
 [homebrew](https://docs.brew.sh/Installation):
 
 ```sh
-brew install bazelisk buildifier hadolint prototool zlib
+brew install bazelisk buildifier clang-format hadolint prototool zlib
 export LDFLAGS="-L/usr/local/opt/zlib/lib"
 export CPPFLAGS="-I/usr/local/opt/zlib/include"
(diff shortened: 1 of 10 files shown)


mostafaelhoushi:
LGTM
Wondering if `demo.py` and `demo_without_bazel.py` can be added to CI? or added to `env_tests.py`?

mostafaelhoushi (approved):

ChrisCummins:
Thanks for the quick response @mostafaelhoushi!

> Wondering if `demo.py` and `demo_without_bazel.py` can be added to CI? or added to `env_tests.py`?

`demo_without_bazel.py` will be tested as part of the CI workflow in the "examples-test" job. Unfortunately `demo.py` (and `env_tests.py`) are not tested in the CI because they are bazel-only :( Long term I will try and de-bazel the example service but that will depend on figuring out a way to build the C++ service without bazel.

Cheers,
Chris

Issue #488, opened 2021-11-17
Circular dependency in the built of the compiler_gym python package


sogartar:
Hi @ChrisCummins,
While porting the build to CMake I encountered a circular dependency problem that I don't know why it does not manifest when running the Bazel build.
There is the following dependency chain.
* `//compiler_gym:compiler_gym`
* `//compiler_gym/envs:envs`
* `//compiler_gym/envs/llvm:llvm`
* `//compiler_gym/envs/llvm:specs`
* `//compiler_gym/envs/llvm/service:compiler_gym-llvm-service-bin`
* `//compiler_gym/envs/llvm/service:compiler_gym-llvm-service-prelinked`
* `//compiler_gym/envs/llvm/service:LlvmSession`
* `//compiler_gym/envs/llvm/service:ActionSpace`
* `//compiler_gym/envs/llvm/service/passes:actions_genfiles`
* `//compiler_gym/envs/llvm/service/passes:make_action_space_genfiles`
* `compiler_gym/envs/llvm/service/passes/make_action_space_genfiles.py`
* `make_action_space_genfiles.py` imports `compiler_gym.envs.llvm.service.passes.common`, which is a part of the `compiler_gym` Python package that has not been built yet.

`compiler_gym/repo/www/requirements.txt` requires `compiler_gym`. The pip package takes precedence, since it has `__init__.py` and hides the partially built `//compiler_gym:compiler_gym`, because it does not have `__init__.py` yet. Even if `compiler_gym/compiler_gym/__init__.py` is copied before that, it depends on the auto-generated `version.py` and the other nested `__init__.py` scripts that depend on the auto-generated gRPC stuff.

I think the proper way to resolve this is all Python scripts that are used for auto-generation during the building of `//compiler_gym:compiler_gym` to be outside of the package. At a first glance these are
* `compiler_gym.envs.llvm.service.passes.common`
* `compiler_gym.envs.llvm.service.passes.config`

They are not used by other parts of the package and are also excluded from the final pip package.

@ChrisCummins, do you think this is a good way to fix the problem? I can include such a change also in the Bazel rules, while doing the CMake change. Or you would like me to first fix it in Bazel as a separate PR?

ChrisCummins:
Hi @sogartar, great detective work!

> I think the proper way to resolve this is all Python scripts that are used for auto-generation during the building of //compiler_gym:compiler_gym to be outside of the package.

[...]

PR #489, opened 2021-11-17
[llvm] Don't use fully-qualified module imports for codegen scripts.

ChrisCummins:
The files compiler_gym/envs/llvm/service/passes/*.py are used to
generate files that are used during the compiler_gym build. Don't use
fully-qualified module paths to emphasize this.

Issue #488.

### DIFF
--- a/compiler_gym/envs/llvm/service/passes/BUILD
+++ b/compiler_gym/envs/llvm/service/passes/BUILD
@@ -1,5 +1,8 @@
 # This package contains scripts for extracting passes from the LLVM source tree
 # and converting them to an action space for reinforcement learning.
+#
+# These scripts are used to programatically generate C++ headers and sources
+# that are then used to compiler the C++ LLVM compiler service.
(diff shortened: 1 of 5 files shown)


ChrisCummins:
CC @sogartar (I can't add you as a reviewer, but feel free to weigh in)

sogartar:
I think this approach should work, but there are two more places that need to be changed.

https://github.com/facebookresearch/CompilerGym/blob/bd32addd51cc1adf89f4ccd411ea1381b5ea2cf6/compiler_gym/envs/llvm/service/passes/config.py#L6
https://github.com/facebookresearch/CompilerGym/blob/6eac3c1b2a5355bb4ed55f243fb834ccfcdd9fcb/compiler_gym/envs/llvm/service/passes/extract_passes_from_llvm_source_tree.py#L36

ChrisCummins:
Good catch, thanks!

PR #493, opened 2021-11-24
[WIP] Trying to Dump IR or Line Numbers of Loops

mostafaelhoushi:
I have been trying to print line numbers of loops, or print the IR content of loops (in order to extract loop features separately) using these statements:
```
      L->getLoopID()->dump();
      L->dump();
      L->getLocRange().getStart().dump();
      L->getLocRange().getEnd().dump();
      L->getStartLoc().getLine();
```

but they all lead to this error:
```
dyld: lazy symbol binding failed: Symbol not found: __ZNK4llvm8Metadata4dumpEv
  Referenced from: /private/var/tmp/_bazel_melhoushi/29d44ffac89bbd569101bd339ac68a0b/execroot/CompilerGym/bazel-out/darwin-fastbuild/bin/examples/example_unrolling_service/loop_unroller/loop_unroller
  Expected in: flat namespace

dyld: Symbol not found: __ZNK4llvm8Metadata4dumpEv
  Referenced from: /private/var/tmp/_bazel_melhoushi/29d44ffac89bbd569101bd339ac68a0b/execroot/CompilerGym/bazel-out/darwin-fastbuild/bin/examples/example_unrolling_service/loop_unroller/loop_unroller
  Expected in: flat namespace
```

I have tried to add `libLLVMRuntimeDyld` as a TARGET to the build system as shown in this pull request but it didn't work. Perhaps I have added it the wrong way. Can @ChrisCummins  please take a look?

### DIFF
--- a/examples/example_unrolling_service/loop_unroller/BUILD
+++ b/examples/example_unrolling_service/loop_unroller/BUILD
@@ -6,18 +6,33 @@
 # This package exposes the LLVM optimization pipeline as a CompilerGym service.
 load("@rules_cc//cc:defs.bzl", "cc_binary")
 
+genrule(
+    name = "libLLVMRuntimeDyld",
+    srcs = [
+        "@clang-llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04//:all_files",
+        "@clang-llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04//:clang",
+    ],
+    outs = ["libLLVMRuntimeDyld.a"],
+    cmd = "cp $$(dirname $(location @clang-llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04//:clang))/../lib/libLLVMRuntimeDyld.a $@",
+    visibility = ["//visibility:public"],
+)
+
 cc_binary(
     name = "loop_unroller",
     srcs = [
         "loop_unroller.cc",
+        #"@clang-llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04//:lib/libLLVMRuntimeDyld.a",
+        ":libLLVMRuntimeDyld",
     ],
     copts = [
         "-Wall",
         "-fdiagnostics-color=always",
         "-fno-rtti",
+        "-force_load",
     ],
     visibility = ["//visibility:public"],
     deps = [
         "@llvm//10.0.0",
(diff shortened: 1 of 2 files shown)


ChrisCummins:
Hey @mostafaelhoushi, I have seen similar errors before. I think the problem is that there is some library missing in the linker flags for LLVM I define here which causes the required `.so` file to not be linked into the binary:

https://github.com/ChrisCummins/bazel_llvm/blob/development/10.0.0/BUILD#L477-L672

To be honest, in the one time I encountered a similar problem, I ended up working around it rather than fixing it, because the debugging loop was so slow 😬

I might get a bit of time towards the start of next week to try and see if I can fix it. If not, is there an alternative API you can call to dump the line numbers?

Cheers,
Chris

ChrisCummins:
@mostafaelhoushi you may find that this kind of error goes away when building with cmake (#498) rather than my janky bazel integration. It isn't yet a full replacement for the bazel build but this may be enough to unblock you.

Cheers,
Chris

mostafaelhoushi:
> @mostafaelhoushi you may find that this kind of error goes away when building with cmake (#498) rather than my janky bazel integration. It isn't yet a full replacement for the bazel build but this may be enough to unblock you.
> 
> Cheers, Chris

Perfect! I will try it with the docker. Thanks Chris!

mostafaelhoushi:
I have good news.
I was able to get CMake working on the loop unroller in this PR: https://github.com/facebookresearch/CompilerGym/pull/530

And then CMake made it easier for me to debug the build error. It turned out that we need to build LLVM in debug mode or add `-DLLVM_ENABLE_DUMP` during the build of LLVM:
https://github.com/llvm/llvm-project/blob/53dc5258282a8322451cba14b5ad2f670028d196/llvm/lib/Analysis/LoopInfo.cpp#L669-L675

image

Instead of having to modify the build of LLVM, I was able to invoke `->print(llvm::dbgs(), true, true)` to print the IR instructions inside the loop. 

So now, we can dump the loop with both bazel and CMake!

I will clean up the code and update this PR.

ChrisCummins:
Fantastic work @mostafaelhoushi! This is a great example of the value of being able to customize the LLVM build that CompilerGym uses. Nice!

cc @sogartar

mostafaelhoushi:
Will close this PR as I am working on another branch to dump loop data

PR #500, opened 2021-12-03
replace the --logfile by --leaderboard_results

yqtian-se:
there is no falg `logfile` defined in the .py file


ChrisCummins:
LGTM. Nice catch @yqtianust and many thanks for the fix!

Cheers,
Chris

PR #518, opened 2021-12-14
[datasets] Make the `site_data_base` path optional.

ChrisCummins:
This allows datasets to be specified without needing to pass a `site_data_base` argument. The purpose of `site_data_base` to provide an on-disk cache for persistent files, but not all datasets need such a cache.


ChrisCummins:
CC @mostafaelhoushi, this resolves one of the TODO notes in your loop environment PR

mostafaelhoushi (approved):
LGTM

PR #520, opened 2021-12-14
Python package building with CMake

sogartar:
Add building of the compiler_gym python package with CMake.

### DIFF
--- a/external/protobuf/build_protobuf.cmake
+++ b/external/protobuf/build_protobuf.cmake
@@ -48,7 +48,16 @@ execute_process(
       "CFLAGS=${CMAKE_C_FLAGS} $ENV{CFLAGS}"
       "CXXFLAGS=-std=c++${CMAKE_CXX_STANDARD} ${CMAKE_CXX_FLAGS} $ENV{CFLAGS}"
       "LDFLAGS=-Wl,-rpath,${CMAKE_INSTALL_PREFIX}/lib ${CMAKE_STATIC_LINKER_FLAGS_INIT} ${CMAKE_SHARED_LINKER_FLAGS_INIT} ${CMAKE_EXE_LINKER_FLAGS_INIT} ${CMAKE_STATIC_LINKER_FLAGS} ${CMAKE_SHARED_LINKER_FLAGS} ${CMAKE_EXE_LINKER_FLAGS} $ENV{LDFLAGS}"
-      ./configure "--prefix=${CMAKE_INSTALL_PREFIX}"
+      ./configure
+        "--prefix=${CMAKE_INSTALL_PREFIX}"
+        # --enable-shared=no seems to not work. It still produces shared libraries.
+        "--enable-shared=no"
+  WORKING_DIRECTORY "${GIT_REPOSITORY_DIR}"
+  COMMAND_ERROR_IS_FATAL ANY)
+
+
+execute_process(
+  COMMAND "${CMAKE_COMMAND}" -E remove_directory "${CMAKE_INSTALL_PREFIX}"
   WORKING_DIRECTORY "${GIT_REPOSITORY_DIR}"
   COMMAND_ERROR_IS_FATAL ANY)
 
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,36 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import argparse
+import distutils.command.build
 import distutils.util
+import fnmatch
 import io
+import sys
+from pathlib import Path
 
 import setuptools
+from setuptools.command.build_py import build_py as build_py_orig
+from setuptools.dist import Distribution
+
+argparser = argparse.ArgumentParser(add_help=False)
+argparser.add_argument(
+    "--package-dir",
+    help="Source directory of package files.",
+    default="bazel-bin/package.runfiles/CompilerGym",
+)
+argparser.add_argument(
+    "--get-wheel-filename",
+    action="store_true",
+    help="Print only output filename without building it.",
+)
+argparser.add_argument(
+    "--build-dir",
+    help="Path to build dir. This is where this script copies files from the source before making the wheel package.",
+    default="build",
+)
+args, unknown = argparser.parse_known_args()
+sys.argv = [sys.argv[0]] + unknown
 
 with open("VERSION") as f:
     version = f.read().strip()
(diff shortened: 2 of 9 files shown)


ChrisCummins:
Hey @sogartar, would you like me to take a look at this or is it still WIP?

Cheers,
Chris

sogartar:
@ChrisCummins, It is ready for review.

setup.py
```diff
@@ -38,78 +60,126 @@ def get_tag(self):
...
+        distutils.command.build.build.initialize_options(self)
+        self.build_base = args.build_dir
+
+
+# Add files that should be excluded from the package.
+# The argument exclude_package_data of setuptools.setup(...)
+# does not work with py files. They have to be excluded here.
+excluded = [
+    str(Path(args.package_dir).joinpath("compiler_gym/envs/llvm/make_specs.py")),
+    str(Path(args.package_dir).joinpath("compiler_gym/bin/random_eval.py")),
```
ChrisCummins:
```suggestion
    str(Path(args.package_dir) / "compiler_gym/envs/llvm/make_specs.py"),
    str(Path(args.package_dir) / "compiler_gym/bin/random_eval.py"),
```

Total bikeshedding, but suggesting `/` operator for consistency with rest of codebase

setup.py
```diff
@@ -5,10 +5,32 @@
...
+    "--package-dir",
+    help="Source directory of package files.",
+    default="bazel-bin/package.runfiles/CompilerGym",
+)
+argparser.add_argument(
+    "--get-wheel-filename",
+    action="store_true",
+    help="Print only output filename without building it.",
+)
+argparser.add_argument("--build-dir", help="Path to build dir.", default="build")
```
ChrisCummins: Could you clarify what is meant by "build" dir in the helpstring? i.e. is it the directory output build files into, the directory for CMake build, etc

sogartar:
@ChrisCummins, I actually encountered some problems with the tests on my environment. I was able to track one of the errors being caused by the `setuptools==59`. The other errors are probably caused by dependencies as well since the tests pass on the CI platform.

I will try to resolve these issues in another PR.

setup.py
```diff
@@ -38,78 +60,126 @@ def get_tag(self):
...
+
+if args.get_wheel_filename:
+    file_name = wheel_name(
+        name="compiler_gym",
+        version=version,
+        cmdclass={"bdist_wheel": bdist_wheel},
+        platforms=[distutils.util.get_platform()],
+    )
+    sys.stdout.write(file_name)
+else:
```
ChrisCummins: This new conditional logic makes the setup.py file quite complicated (at least to me, without being familiar with the intent). Could you add a comment giving a high-level overview of what all this code is for? Maybe at the top of the file, or wherever feels best

Makefile
```diff
@@ -197,7 +197,7 @@ bazel-build: bazel-fetch
 	$(BAZEL) $(BAZEL_OPTS) build $(BAZEL_BUILD_OPTS) $(BUILD_TARGET)
 
 bdist_wheel: bazel-build
-	$(PYTHON) setup.py bdist_wheel
+	$(PYTHON) setup.py bdist_wheel --package-dir bazel-bin/package.runfiles/CompilerGym
```
ChrisCummins: Really minor nitpick: no need to set this flag as it is the default value.

external/protobuf/build_protobuf.cmake
```diff
@@ -48,7 +48,16 @@ execute_process(
       "CFLAGS=${CMAKE_C_FLAGS} $ENV{CFLAGS}"
       "CXXFLAGS=-std=c++${CMAKE_CXX_STANDARD} ${CMAKE_CXX_FLAGS} $ENV{CFLAGS}"
       "LDFLAGS=-Wl,-rpath,${CMAKE_INSTALL_PREFIX}/lib ${CMAKE_STATIC_LINKER_FLAGS_INIT} ${CMAKE_SHARED_LINKER_FLAGS_INIT} ${CMAKE_EXE_LINKER_FLAGS_INIT} ${CMAKE_STATIC_LINKER_FLAGS} ${CMAKE_SHARED_LINKER_FLAGS} ${CMAKE_EXE_LINKER_FLAGS} $ENV{LDFLAGS}"
-      ./configure "--prefix=${CMAKE_INSTALL_PREFIX}"
+      ./configure
+        "--prefix=${CMAKE_INSTALL_PREFIX}"
+        # --enable-shared=no seems to not work.
```
ChrisCummins: Not sure what this means

ChrisCummins (approved):
Hi @sogartar, I haven't tested this locally but I have read through the PR and it looks great, thank you! I think this is ready to merge, modulo some minor comments that I have left inline. 

I like that you are using the same `make install-test` target for both CMake and bazel. This should make debugging a lot easier during the transition.

Cheers,
Chris

setup.py
sogartar: Done.

setup.py
sogartar: Done.

setup.py
sogartar: I added a comment.
The goal here is to be able to query the script for its byproducts. `setuptools` does not have an out-of-the-box facility for that. Then you can use this result in the build when describing the output of the target.

Makefile
sogartar: Done.

external/protobuf/build_protobuf.cmake
sogartar: I improved the comment.

sogartar:
@ChrisCummins, why does the package include `libLLVMPolly.so`? It looks like it is not used anywhere.

sogartar:
@ChrisCummins, I force pushed my last changes. I did not know that Github did not support a good force push, rebase workflow. It may be difficult to compare what has changed.
In my fork I tagged the previous version to `cmake-py-pkg-pr-1`.

setup.py
ChrisCummins: Ah I see, that's clever :)

ChrisCummins:
Thanks @sogartar! Yeah it can be a bit hard to find the changes when you force push, not too bad in this case though :)

> @ChrisCummins, why does the package include `libLLVMPolly.so`? It looks like it is not used anywhere.

Purely a workaround for a weird bug in the bazel LLVM build where the compiled binaries try to load libLLVMPolly.so at launch, even though there's no use of the Polly APIs. Since your CMake build doesn't have this problem we can completely drop libLLVMPolly.so and the associated `patchelf` stuff like [here](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/BUILD#L29-L46).

Cheers,
Chris

ChrisCummins (approved):
LGTM, merging. Thanks @sogartar!

Cheers,
Chris

mostafaelhoushi:
Please review this PR, I had to add a few more instructions when I built using CMake:
https://github.com/facebookresearch/CompilerGym/pull/529

PR #529, opened 2021-12-25
Update CMake Instructions

mostafaelhoushi:
I followed the CMake instructions on a newly installed Ubuntu virtual machine (using the latest Ubuntu version out there).
This PR has minor changes required for me to get it to work

The main addition I made is the command to install .whl file.
There are also some minor changes that you may feel are not necessary or may break things on other versions of Linux, so I can remove them if you want.

On another note, I feel that the conda environment we create is necessary and shouldn't be removed. At least that what I noticed for both bazel and CMake builds, because we need the libraries installed by the `make init` command to stay in the environment.
(You may have noticed I have added a `--force-reinstall` when installing the .whl file in order to overwrite the public `compiler_gym` library installed by `make init`)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -98,7 +98,7 @@ Darwin is not supported with CMake.
 ### Dependency instructions for Ubuntu
 
 ```bash
-sudo apt-get install lld-9 \
+sudo apt-get install g++ lld-9 \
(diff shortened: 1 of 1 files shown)


ChrisCummins:
LGTM. Thanks @mostafaelhoushi!

Cheers,
Chris

PR #530, opened 2021-12-26
Example unrolling service without bazel

mostafaelhoushi:
Here I build the `loop_unroller` tool within the repo's CMake. Do you think it is better to have a separate/isolated build script for it?

### DIFF
--- a/examples/example_unrolling_service/example_without_bazel.py
+++ b/examples/example_unrolling_service/example_without_bazel.py
@@ -1,0 +1,174 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""This module demonstrates how to """
+import os
+import subprocess
+from pathlib import Path
+from typing import Iterable
+
+import compiler_gym
+from compiler_gym.datasets import Benchmark, Dataset
+from compiler_gym.envs.llvm.llvm_benchmark import get_system_includes
+from compiler_gym.spaces import Reward
+from compiler_gym.third_party import llvm
+from compiler_gym.util.registration import register
+
+UNROLLING_PY_SERVICE_BINARY: Path = Path(
+    "example_unrolling_service/service_py/example_service.py"
+)
+assert UNROLLING_PY_SERVICE_BINARY.is_file(), "Service script not found"
+
+BENCHMARKS_PATH: Path = Path("example_unrolling_service/benchmarks")
+
+NEURO_VECTORIZER_HEADER: Path = Path(
+    "../compiler_gym/third_party/neuro-vectorizer/header.h"
+)
+
+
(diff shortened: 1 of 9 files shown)


examples/example_unrolling_service/loop_unroller/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+llvm_map_components_to_libnames(_LLVM_LIBS analysis core irreader support passes)
+cg_cc_binary(
+  NAME
+    loop_unroller
+  SRCS
+    "loop_unroller.cc"
+  COPTS
+    "-Wall"
+    "-fdiagnostics-color=always"
```
ChrisCummins: Nitpick: are these necessary?

ChrisCummins (approved):
LGTM, thanks @mostafaelhoushi!

I think the approach of adding a new build option for the examples directory makes sense to me. @sogartar, as the author of the CMake build, does that look good to you?

Could you update the CI job to build with -DCOMPILER_GYM_BUILD_EXAMPLES=on?

examples/example_unrolling_service/loop_unroller/CMakeLists.txt
mostafaelhoushi: Removing it

examples/example_unrolling_service/example_without_bazel.py
```diff
@@ -0,0 +1,188 @@
...
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+"""This script demonstrates how the Python example service without needing
+to use the bazel build system.
+
+Prerequisite:
+    # In the repo's INSTALL.md, follow the 'Building from source using CMake' instructions with `-DCOMPILER_GYM_BUILD_EXAMPLES=ON` added to the `cmake` command
+    # Then copy the `loop_unroller` binary
+    $ cd <path to source directory>/examples
+    $ cp <path to build directory>/examples/example_unrolling_service/loop_unroller/loop_unroller ./example_unrolling_service/loop_unroller/
```
mostafaelhoushi: @ChrisCummins here as a hack I am copying the executable from the build directory to next to the Python file in the source directory. 
What should be the proper way of doing this? Creating a `py_binary` for the Python script in CMake and adding the executable as a dependency? I would prefer a way that would treat the script as a standalone.

Maybe add the path to the executable to the `PATH` environment variable?

sogartar: If you make a `py_binary` it will create a symlink in the build directory under the same subdirectory as is the source file relative to the top source directory.
Then you run the symlink.
This is generally the correct aporach.

sogartar:
@ChrisCummins, I missed your question.
Adding `COMPILER_GYM_BUILD_EXAMPLES` to the CMake cache is good.

Issue #524, opened 2021-12-21
Switch benchmark URI parser to urllib.parse()


ChrisCummins:
## 🚀 Feature

Python already has a [great parser](https://docs.python.org/3/library/urllib.parse.html) for URIs. Use it instead of these ad-hoc regexes:

https://github.com/facebookresearch/CompilerGym/blob/398b39f4cda06f43ee119c9b488aaf7b45266f79/compiler_gym/datasets/uri.py#L8-L22

## Motivation

The goal is to lay the groundwork to support more flexible benchmark URIs, such as those with query parameters: `benchmark://cbench-v1/qsort?dataset=1`.

## Pitch

Swap out the ad-hoc regexes and parsers for the builtin `urllib.parse.urlparse()` function. This should not affect the behavior of any of the existing datasets, but it may allow us to be more lenient in accepting "unusual" URIs, and new features like query parameters.

PR #555, opened 2022-01-27
[datasets] Small benchmark URI fix and tweak

ChrisCummins:
This fixes a typo and test regression introduced in #524.


hughleat (approved):
LGTM

If this string processing is frequent and creating then unparsing is expensive, consider caching `__repr__`?

ChrisCummins:
Re. Caching `__repr__`, good idea! I'll do this

PR #563, opened 2022-02-05
Update statistics.py

### LINKED ISSUE #562: `statistics.py` wrong parameter name
## 🐛 Bug

The functions [here](https://github.com/facebookresearch/CompilerGym/blob/e248330d2475fbcdf473cc3df951f25b5eaf4945/compiler_gym/util/statistics.py#L8) says they take `iterable` as inputs. However, `np.asarray` actually take `array_like`.

[Quote:
](https://numpy.org/doc/stable/reference/generated/numpy.asarray.html)

> Input data, in any form that can be converted to an array. This includes lists, lists of tuples, tuples, tuples of tuples, tuples of lists and ndarrays.

e.g.
```python
geometric_mean(i for i in range(10))
```
This will fail because though it's an `iterable`, it's not an `array_like`.

uduse:
Fixes #562

### DIFF
--- a/compiler_gym/util/statistics.py
+++ b/compiler_gym/util/statistics.py
@@ -5,9 +5,9 @@
 import numpy as np
 
 
-def geometric_mean(iterable):
+def geometric_mean(array_like):
     """Zero-length-safe geometric mean."""
-    values = np.asarray(iterable)
+    values = np.asarray(array_like)
     if not values.size:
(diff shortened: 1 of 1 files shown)


ChrisCummins:
Good catch, thanks for the patch @uduse! Merging.

Cheers,
Chris

PR #567, opened 2022-02-08
[cmake] Handful of small tweaks

ChrisCummins:
This patch set:

- Removes the "subprocess" library from the bazel and CMake build configs, as it is unused as of #398.
- Fixes a typo `s/string/STRING/` that caused a harmless error to be printed.
- Removes Polly from the CMake build config. It is not used by the LLVM environment. It is only included in the Bazel build only as a workaround for a weird bug in my Bazel LLVM config.
- Strips the `include/` prefix from all LLVM header includes, and the workaround in the CMake config.
- Improves the ad-hoc code for parsing LLVM sources to extract the transform pass names, fixing a couple of bugs I discovered while building against LLVM 13.0.0.1.


ChrisCummins:
cc @sogartar (I can't add you as a reviewer)

sogartar (approved):
I don't have any remarks. Since you mention that `parse_initialize_pass` is a "throw-away" code, I was not thorough with its review.

PR #569, opened 2022-02-10
Add a CMake formatter and linter to toolchain.

ChrisCummins:
CompilerGym uses pre-commit hooks to automate code formatting and run a handful of linters (cf [CONTRIBUTING.md](https://github.com/facebookresearch/CompilerGym/blob/development/CONTRIBUTING.md#code-style)). For example, we use black for Python sources, clang-format for C++, and buildifier for Bazel BUILD files. Now that there is a significant quantity of CMakeLists.txt files, we should do the same for CMake.

This adds [cmakelang](https://github.com/cheshirekow/cmake_format) formatter and linter to the pre-commit checks. I configured the code formatter to use a style similar to Python: 4 space indentation, 100 character line wrapping and dangling parentheses. AN 

Here's a random sample of what it looks:

```cmake
if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
    message(STATUS "No build type selected, default to ${DEFAULT_CMAKE_BUILD_TYPE}")
    set(CMAKE_BUILD_TYPE
        "${DEFAULT_CMAKE_BUILD_TYPE}"
        CACHE STRING "Build type (default ${DEFAULT_CMAKE_BUILD_TYPE})" FORCE
    )
endif()
```

One thing I don't like is that it flattens the indentation for all function arguments, which makes the argument lists to IREE's bazel macros a bit harder to read:

```diff
cg_genrule(
-  NAME
+    NAME
     libLLVMPolly
-  OUTS
+    OUTS
     "libLLVMPolly.so"
-  COMMAND
+    COMMAND
     "cp $<TARGET_FILE:LLVMPolly> $@"
     ABS_DEPENDS
     LLVMPolly
)
```

However I don't see this as a total deal breaker as we should think about replacing those macros with more idiomatic CMake configs. I'm open to suggestions if there are configuration options that people would prefer. You can read about configuring cmakelang [here](https://cmake-format.readthedocs.io/en/latest/configuration.html).

P.S: This PR only adds the config to run the formatter and linter. Once accepted, I will run the formatter on all of the existing sources and commit the changes, but this will make a very large diff.


ChrisCummins:
@sogartar, @@KyleHerndon, happy to hear your thoughts on this, and if you have large pending changes to the CMakeLists files then let me know as I can hold off running the formatter to prevent a nasty merge conflict.

Cheers,
Chris

ChrisCummins:
You can see the diff that would be applied by this code formatter by looking at the output of the Pre-commit/all CI job.

Cheers,
Chris

sogartar:
It would be nice to have a formatter for CMake.
Looking at the [issues](https://github.com/cheshirekow/cmake_format/issues) of this formatter, it seems this is not actively developed and also it is not mature.
For example these issues:
https://github.com/cheshirekow/cmake_format/issues/283
https://github.com/cheshirekow/cmake_format/issues/279
https://github.com/cheshirekow/cmake_format/issues/272
https://github.com/cheshirekow/cmake_format/issues/268
https://github.com/cheshirekow/cmake_format/issues/264

If we use it we may have to become a contributer.

ChrisCummins:
Ah that's a shame. I did a bit of browsing around and found this:
https://github.com/BlankSpruce/gersemi

It looks like a much younger project but is more actively developed, and the code examples they provide show nicely indented parameter groups :) I'll have a play around with it.

Cheers,
Chris

ChrisCummins:
I have played around with https://github.com/BlankSpruce/gersemi and it seems to do a good job on our CMake build of making sure trivial things like indentation and whitespace are consistent. I'll merge it in now.

Cheers,
Chris

PR #531, opened 2021-12-28
gRPC refactoring of actions and observations

sogartar:
Fixes #526

This is a draft of the protobuf message definitions. It makes actions and observations be more like the CompilerGym Python environment interface.

@ChrisCummins, could you take a look and see if this structure is appropriate. After I address your remarks I will proceed to refactor the existing environments to use the new structure.

PR #570, opened 2022-02-14
Add Observations to Loop Optimizations Environment

mostafaelhoushi:
- Adds inst2vec [Done], Autophase [Done], and Programl [In Progress] to the Loop Optimizations Environment
- to enable extracting `ProGraml` I created a small `compute_programl.cc` program. I tried first to place it in `external/programl` but was getting build errors, so I temporarily placed it in the loop opt example directory. 

Issues:
- we need to first buld `compute_autophase` separately before we can run our example: 
```
bazel build //compiler_gym/third_party/autophase:compute_autophase-prelinked

bazel run //examples/loop_optimizations_service:example
```

TODO (before merging):
- [x] fix build scripts so that we won't need to build `compute_autophase` 
- [x] move `compute_programl` to a suitable directory
- [x] complete Programl observation
- [x] add unit test cases for each observation
- [x] add AutoPhase dictionary observation
- [ ] verify build with CMake on Linux


examples/loop_optimizations_service/service_py/BUILD
```diff
@@ -8,21 +8,36 @@ py_binary(
...
+        "//compiler_gym/third_party/inst2vec",
         "//compiler_gym/third_party/llvm",
     ],
 )
 
-py_library(
-    name = "utils",
-    srcs = ["utils.py"],
+cc_binary(
+    name = "compute_programl-prelinked",
```
ChrisCummins:
```suggestion
    name = "compute_programl",
```

The "-prelinked" suffix is used for a couple of targets where we need to do some post-compile trickery using `patchelf` to resolve the libLLVMPolly.so library ([e.g.](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/BUILD#L29-L46)). IUC, that isn't needed here so the target can just be called "compute_programl".

examples/loop_optimizations_service/service_py/loops_opt_service.py
```diff
@@ -223,11 +237,21 @@ def get_observation(self, observation_space: ObservationSpace) -> Event:
...
+        elif observation_space.name == "Inst2vec":
+            Inst2vec_str = self.inst2vec.preprocess(self.ir)
+            Inst2vec_ids = self.inst2vec.encode(Inst2vec_str)
+            return Event(
+                int64_tensor=Int64Tensor(shape=[len(Inst2vec_ids)], value=Inst2vec_ids)
+            )
+        elif observation_space.name == "Autophase":
+            Autophase_str = run_command(
+                [
+                    "../../../compiler_gym/third_party/autophase/compute_autophase-prelinked",
```
ChrisCummins: Strongly advise against relpaths. Use [runfiles_path](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/util/runfiles_path.py#L22-L55) instead.

examples/loop_optimizations_service/service_py/BUILD
mostafaelhoushi: Done

mostafaelhoushi:
@ChrisCummins, I addressed the comments and enabled Programl observation (also moved `compute_programl` to `compiler_gym/third_party/programl`.

However, I am currently printing programl observation as a Json string. Is that how the originl CompilerGym LLVM environments print Programl observation?

ChrisCummins:
> @ChrisCummins, I addressed the comments and enabled Programl observation (also moved `compute_programl` to `compiler_gym/third_party/programl`.

Thanks! And yep, `compiler_gym/third_party` sounds like a good idea.

> However, I am currently printing programl observation as a Json string. Is that how the originl CompilerGym LLVM environments print Programl observation?

This is exactly what the LLVM environment does:

https://github.com/facebookresearch/CompilerGym/blob/11784d436fc9bd4c3d9a1588508a970649ac5ed3/compiler_gym/envs/llvm/service/Observation.cc#L101-L111

However, since merging #531, this could be made more efficient for the LLVM environment by embedding the ProgramGraph protocol buffer directly in the observation, removing the need to do a JSON serialize/deserialize. I haven't got around to doing that refactor yet though.

Cheers,
Chris

ChrisCummins (approved):
I agree with your TODO list in the PR message about adding some extra test cases to cover these new observations. Otherwise, LGTM!

examples/loop_optimizations_service/service_py/loops_opt_service.py
```diff
@@ -176,7 +214,9 @@ def apply_action(self, action: Event) -> Tuple[bool, Optional[ActionSpace], bool
                 args[i] = arg
             run_command(
                 [
-                    "../opt_loops/opt_loops",
+                    runfiles_path(
+                        "examples/loop_optimizations_service/opt_loops/opt_loops"
+                    ),
```
mostafaelhoushi: @ChrisCummins I have changed it to `runfiles_path` and it is working fine with bazel, but with CMake it is not working... I have added a CMakeLists.txt file to make `loops_opt_service.py` dependent on `opt_loops` but it is still not working. Do you have an idea how to get it working with CMake?

ChrisCummins: Oof. Erm, I have no idea how data dependencies work in CMake, sorry! @sogartar?

ChrisCummins: If the file exists, but resolving the path fails, then you could ditch `runfiles_path()` in favor of something like `Path(os.path.join(os.path.dirname(__file__), "path/to/my/thing")).resolve()` to get an absolute path to the dependency.

The only reason I suggested not using relpaths is in case the user `os.chdir()`s.

ChrisCummins: The whole `runfiles_path()` mechanism will need a total overhaul when migrating fully to CMake build.

sogartar: Runfiles depend on Bazel if the environment variable `COMPILER_GYM_RUNFILES` is not defined. With CMake the Bazel resolution will probably fail, so it falls back to resolving the path as a relative path with respect to `_PACKAGE_ROOT`. Either you define the `COMPILER_GYM_RUNFILES` env var or have the file in the exact relative location with respect to `_PACKAGE_ROOT`.

sogartar: Another problem may be that the file does not get where it should be. This means that something is wrong with the dependencies and the target that makes the files is not triggered. There may be a bug in `cg_add_data_dependencies(...)`.
Another problem may come from `cg_genrule` resolving to the wrong command, but this should generate an error when executed.
@mostafaelhoushi, you can run the CMake configuration with `--trace-expand` flag and see what happens line-by-line.
You can also use CMake's `--graphviz` option to visualize the dependency graph.

ChrisCummins: Thanks for the `--graphviz` tip, cool!

mostafaelhoushi: Thanks @sogartar for your guidance. I tried to debug it and I think this is the cause:
>> something is wrong with the dependencies and the target that makes the files is not triggered

For now I used the `Path(os.path.join(os.path.dirname(__file__), "path/to/my/thing")).resolve()` workaround and created this issue #603

examples/loop_optimizations_service/service_py/BUILD
```diff
@@ -8,21 +8,19 @@ py_binary(
     name = "loops-opt-service-py",
     srcs = ["loops_opt_service.py"],
     data = [
+        "//compiler_gym/third_party/autophase:compute_autophase-prelinked",
+        "//compiler_gym/third_party/programl:compute_programl",
```
mostafaelhoushi, line 12: Building on Linux is failing as I get this error (see CI log [here](https://github.com/facebookresearch/CompilerGym/runs/5417482434?check_suite_focus=true)):
```
-- Configuring done
CMake Error at build_tools/cmake/cg_macros.cmake:281 (add_dependencies):
  The dependency target
  "compiler_gym__third_party__autophase_compute_autophase" of target
  "examples__loop_optimizations_service__service_py__loops_opt_service" does
  not exist.
Call Stack (most recent call first):
  build_tools/cmake/cg_py_library.cmake:78 (cg_add_data_dependencies)
  examples/loop_optimizations_service/service_py/CMakeLists.txt:8 (cg_py_library)


CMake Error at build_tools/cmake/cg_macros.cmake:281 (add_dependencies):
  The dependency target
  "compiler_gym__third_party__programl_compute_programl" of target
  "examples__loop_optimizations_service__service_py__loops_opt_service" does
  not exist.
Call Stack (most recent call first):
  build_tools/cmake/cg_py_library.cmake:78 (cg_add_data_dependencies)
  examples/loop_optimizations_service/service_py/CMakeLists.txt:8 (cg_py_library)


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
Error: Process completed with exit code 1.
```

I tried many times to fix it, and with the issue that CMake configuration takes multiple hours (see #595 ), it was really difficult to solve it. So I merged for now and will fix it later.

Cc @ChrisCummins @sogartar

examples/loop_optimizations_service/service_py/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+
+cg_py_library(
+  NAME
+    loops_opt_service
+  SRCS
+    "loops_opt_service.py"
+  DATA
+    examples::loop_optimizations_service::opt_loops::opt_loops
+    compiler_gym::third_party::autophase:compute_autophase
+    compiler_gym::third_party::programl:compute_programl
```
mostafaelhoushi, line 16: Building on Linux is failing as I get this error (see CI log [here](https://github.com/facebookresearch/CompilerGym/runs/5417482434?check_suite_focus=true)):
```
-- Configuring done
CMake Error at build_tools/cmake/cg_macros.cmake:281 (add_dependencies):
  The dependency target
  "compiler_gym__third_party__autophase_compute_autophase" of target
  "examples__loop_optimizations_service__service_py__loops_opt_service" does
  not exist.
Call Stack (most recent call first):
  build_tools/cmake/cg_py_library.cmake:78 (cg_add_data_dependencies)
  examples/loop_optimizations_service/service_py/CMakeLists.txt:8 (cg_py_library)


CMake Error at build_tools/cmake/cg_macros.cmake:281 (add_dependencies):
  The dependency target
  "compiler_gym__third_party__programl_compute_programl" of target
  "examples__loop_optimizations_service__service_py__loops_opt_service" does
  not exist.
Call Stack (most recent call first):
  build_tools/cmake/cg_py_library.cmake:78 (cg_add_data_dependencies)
  examples/loop_optimizations_service/service_py/CMakeLists.txt:8 (cg_py_library)


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
Error: Process completed with exit code 1.
```

I tried many times to fix it, and with the issue that CMake configuration takes multiple hours (see #595 ), it was really difficult to solve it. So I merged for now and will fix it later.

Cc @ChrisCummins @sogartar

PR #574, opened 2022-02-16
[service] Loosen gRPC constraints on message sizes.

ChrisCummins:
Remove the size limit on incoming and outgoing messages, and
increase the mazimum allowed metadata message size.

This is all to improve the reliability of the service when working
with particuluarly large programs / observations, that can exceed
the quite tight default resource constraints in gRPC.


mostafaelhoushi (approved):
LGTM

PR #575, opened 2022-02-17
[llvm] Add a method to compute .text size of the compiled binary

ChrisCummins:
The `ObjectTextSize` observation space currently does the equivalent of:

```
clang -c input.bc -o foo.o
llvm-size foo.o
```

This fine for the "basic" case, but it is totally inflexible. What if the user wants to pass in a "-D" define flag? Compile for a different architecture? Control the optimizations applied to the backend?

To address these, this PR adds a mechanism by which users can pass in a custom compilation command (e.g. `$CC -target foobar -Oz $IN`) which the environment can run to then compute the .text section size.

Summary of changes:

- Pull out the "RealizedBenchmarkDynamicConfig" class into its own header and rename it "BenchmarkDynamicConfig" for consistency.

- Add a new `getTextSizeInBytes()` overload that takes a "BenchmarkDynamicConfig" as input and runs the build command to produce a binary.

- Expose the above function through a new set of binary text size observation spaces: `TextSizeBytes`, `TextSizeO0`, `TextSizeO3`, `TextSizeOz`. Long term, these should be merged with "ObjectTextSize" spaces, but this will change the semantics of the ObjextTextSize space so keep them separate for now.

As a side note, the relationship between the cost functions and Benchmark class are a total mess, the whole thing is due an overhaul. However, that is beyond the scope of this PR. This PR attempts only to bolt on some new functionality needed to support cross-compiling.


mostafaelhoushi (approved):
LGTM

Issue #325, opened 2021-07-20
Extend RPC interface to support multi-file benchmarks


ChrisCummins:
## 🚀 Feature

Add support for benchmarks which comprise multiple files.

## Motivation

Presently the `Benchmark` protobuf requires that each benchmark is a single file (called the "program"):

```proto
// Representation of the input to a compiler.
message Benchmark {
  // The name of the benchmark to add. In case of conflict with an existing
  // benchmark, this new benchmark replaces the existing one.
  string uri = 1;
  // The description of the program that is being compiled. It is up to the
  // service to determine how to interpret this file, and it is the
  // responsibility of the client to ensure that it provides the correct format.
  // For example, the service could expect that this file contains serialized
  // IR data, or an input source file.
  File program = 2;
  // An optional configuration option that details how to build and run the
  // benchmark program.
  BenchmarkDynamicConfig dynamic_config = 3;
}
```

In the LLVM environment, we presently support multi-file benchmarks by linking all of the source files into a single LLVM module to construct a benchmark. This means that we can't support benchmarks for which the module that we are optimizing is only a single component of a larger program.

## Pitch

Replace the single file "program" field with a new "files" mapping which maps file names to file contents, and a "main" field which tells the compiler which input is the "main" one (if applicable):

```proto
// Representation of the inputs to a compiler.
message Benchmark {
  // The name of the benchmark to add. In case of conflict with an existing
  // benchmark, this new benchmark replaces the existing one.
  string uri = 1;
  // A mapping from file name to file contents for each of the components that
  // makes up this benchmark. It is up to the service to determine how to
  // interpret these files, and it is the responsibility of the client to ensure
  // that it provides the correct format. For example, the service could expect
  // that each file contains serialized IR data, or an input source file.
  map<string, bytes> files = 4;
  // A key into the "files" mapping that identifies the "main" file for a
  // benchmark, if applicable.
  string main = 5;
  // An optional configuration option that details how to build and run the
  // benchmark program.
  BenchmarkDynamicConfig dynamic_config = 4;
  // Deprecated fields:
  reserved 3;
  reserved "program";
}
```

For example:

```
Benchmark {
  uri = "benchmark://gcc-v0/my-app"
  files {
    "src/main.c": "..."
    "include/header.h": "..."
    "src/util.cc": "..."
  }
  main = "src/main.c"
}
```

[...]

PR #577, opened 2022-02-17
[llvm] Add API to construct a benchmark from clang invocation.                                                                                                                                                                                                                  

ChrisCummins:
This adds a new API for constructing a benchmark from a command line invocation of a compiler. For example, given the command line:

```sh
$ gcc a.c src/b.c -DNDEBUG -lm -Oz -o foo.exe
```

The `make_benchmark_from_command_line()` method interprets this command line and rewrites it to one that emits an LLVM-IR file:

```
/path/to/compilergym/clang a.c src/b.c -DNDEG -lm -Oz -emit-llvm -c -o /path/to/benchmark.bc -Xclang disable-llvm-optmzns
```

It then loads the resulting bitcode for use as a benchmark. When the user is done playing around with the benchmark, they call a special `compiler()` method on the benchmark which re-invokes the original command line, but this time using the bitcode as input:

```
/path/to/compilergym/clang /path/to/benchmark.nc -DNDEG -lm -Oz -o foo.exe
```

Putting this all together, here is an end-to-end example using CompilerGym as a substitute for LLVM's pipeline:

```py
>>> import compiler_gym
>>> env = compiler_gym.make("llvm-v0")
>>> benchmark = env.make_benchmark_from_command_line(                                                                                                                                                                                                                       
...    "gcc a.c src/b.c -DNDEBUG -lm -Oz -o foo.exe"
... )                                                                                                                                                                                                                                                                       
>>> for _ in range(100):
...     _, _, done, _ env.step(env.action_space.sample()
...     if done:
...         env.reset()
>>> env.benchmark.compile()
# Now the file foo.exe is compiled
```
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
The idea is to make it easier to integrate CompilerGym into an existing build system, as it is normally possible to dump a verbose log of all of the compiler invocations run during a build.


compiler_gym/envs/llvm/benchmark_from_command_line.py
```diff
@@ -0,0 +1,75 @@
...
+from compiler_gym.util.runfiles_path import transient_cache_path
+
+logger = logging.getLogger(__name__)
+
+
+class BenchmarkFromCommandLine(Benchmark):
+    def __init__(self, command_line: List[str], bitcode: bytes, timeout: int):
+        uri = (
+            f"benchmark://clang-v0/{urllib.parse.quote_plus(shlex.join(command_line))}"
+        )
```
mostafaelhoushi: The dataset or benchmark needs to be named `clang-v0`?

compiler_gym/envs/llvm/llvm_env.py
```diff
@@ -581,3 +592,101 @@ def fork(self):
...
+        """Create a benchmark for use with this environment.
+
+        This function takes a command line compiler invocation as input,
+        modifies it to produce an unoptimized LLVM-IR bitcode, and then runs
+        the modified command line to produce a bitcode benchmark.
+
+        For example, the command line:
+
+            >>> benchmark = env.make_benchmark_from_command_line(
+            ...     ["gcc", "-DNDEBUG", "a.c", "b.c", "-o", "foo", "-lm"]
```
mostafaelhoushi, line 658: I am wondering if "gcc" works here... as down below we add `-Xclang` to the arguments

compiler_gym/third_party/gccinvocation/gccinvocation.py
```diff
@@ -0,0 +1,210 @@
...
+            "c89",
+            "c99",
+            "cc",
+            "gcc",
+            "c++",
+            "g++",
+            "xgcc",
+            "clang",
+            "clang++",
+        )
```
mostafaelhoushi, line 95: so is this library literary `gccinvokcation` or is it also a generic C compiler invocator?

mostafaelhoushi (commented):
Super Cool!
I am wondering is that environment compatible with `LlvmEnv` or `GccEnv`? I mean can we directly call the existing observation or action functions on that environment?

mostafaelhoushi (approved):

compiler_gym/third_party/gccinvocation/gccinvocation.py
ChrisCummins: I've hacked on it to add support for clang by adding support for a few clang-specific flags, though I'm sure I've missed some

ChrisCummins: I don't want to change the name of the third-party library, though if my edits work well then I could submit them upstream and consider renaming it to "ccinvocation" or something

compiler_gym/envs/llvm/llvm_env.py
ChrisCummins: Yes, this function takes an argument `replace_driver` (default=True) that swaps out whatever the first argument is for clang. So this will work:

```py
>>> env.make_benchmark_from_command_line(["gcc", "in.c"])
```

and will get translated to something like:

```
/path/to/compiler_gym/llvm/bin/clang in.c -c -emit-llvm -o - -Xclang disable-llvm-optmzns
```

but this will not:

```py
>>> env.make_benchmark_from_command_line(["gcc", "in.c"], replace_driver=False)
```

The idea with `replace_driver=False` is that it allows you to use your own compiler to produce the input bitcode, providing your compiler produces compatible bitcode.

The downside is that it won't work with command line options that aren't the same between whatever compiler the user uses and our version of clang.

ChrisCummins: I'll extend the docstring to explain this better

hughleat (approved):
LGTM.
Quick question though. How does it deal with mixed .c and .o files? E.g.:
```bash
gcc foo.c bar.o -o foo
```
My guess is that bar.o will be considered a source and that clang will barf trying to bitcode-ise it. Should those be filtered out for linking?

ChrisCummins:
> LGTM. Quick question though. How does it deal with mixed .c and .o files?

Good question! I added support so that `.o` files are excluded from the bitcode environment, but are kept for linking later.

The current implementation for this will not work if the CompilerGym backend and frontend are on different machines. To fix that, we need #325.

Cheers,
Chris

Issue #572, opened 2022-02-16
E0216 from backup_poller.cc or ev_epollex_linux.cc


uduse:
## 🐛 Bug

This happens when I spawn multiple compiler gym sessions in parallel. The error comes in two forms but I think they similar:

either:
```
E0216 05:27:23.400729219  198153 ev_epollex_linux.cc:515]    Error shutting down fd 16. errno: 9```
```
or:
```
E0216 05:27:43.469307484  293529 backup_poller.cc:134]       Run client channel backup poller: {"created":"@1645007263.469241393","description":"pollset_work","file":"src/core/lib/iomgr/ev_epollex_linux.cc","file_line":320,"referenced_errors":[{"created":"@1645007263.469233691","description":"Bad file descriptor","errno":9,"file":"src/core/lib/iomgr/ev_epollex_linux.cc","file_line":950,"os_error":"Bad file descriptor","syscall":"epoll_wait"}]}
```

## To Reproduce

I managed to create this example to consistently recreates this behavior:

```python
import compiler_gym
import ray


@ray.remote
def spawn():
    with compiler_gym.make("llvm-ic-v0") as env:
        env.reset()
        _, _, done, _ = env.step(env.action_space.sample())
        if done:
            raise RuntimeError


ray.get([spawn.remote() for _ in range(5000)])
```

example output from the above script:
```
$ python sandbox.py 
2022-02-16 05:31:32,323 INFO services.py:1374 -- View the Ray dashboard at http://127.0.0.1:8265
(spawn pid=438029) E0216 05:31:43.796126540  482891 backup_poller.cc:134]       Run client channel backup poller: {"created":"@1645007503.796063126","description":"pollset_work","file":"src/core/lib/iomgr/ev_epollex_linux.cc","file_line":320,"referenced_errors":[{"created":"@1645007503.796054426","description":"Bad file descriptor","errno":9,"file":"src/core/lib/iomgr/ev_epollex_linux.cc","file_line":950,"os_error":"Bad file descriptor","syscall":"epoll_wait"}]}
[...]

PR #578, opened 2022-02-17
Hardening patchset

ChrisCummins:
This patch set contains a handful of fixes for issues discovered while trying to track down #572:

- Corrects a misuse of the boost::process API, fixing a bug wherein llvm-size child processes could become detached from the parent environment service and be left as zombie processes.
- Disables TCP port reuse for all gRPC channels, as it has been reported to occasionally interfere with port autoselect (see https://github.com/grpc/grpc/issues/10755).
- Ensures that all gRPC client/server connection options in both Python and C++ use the same configuration options.
- Fix a bug in which connection liveness was reported exactly the wrong way around (dead services were logged as alive and alive connections as dead).
- Replaces `get_system_includes()` with `get_system_library_flags()`, a function that returns a list of `-isystem` flags, and a `-L` flag for macOS. This **enables optimizing for runtime on macOS**.
- Holds the gRPC version back below 1.44 until I can diagnose the new error messages raised in the new version. See #572.
- Adds a 300 second timeout to all tests when using the pytest test runner.


mostafaelhoushi (approved):
LGTM

PR #594, opened 2022-03-03
[third-party] Mark inst2vec regexes as regex literals.

ChrisCummins:
It is now deprecated to use regex escape characters in string literals.

This adds the "r" string prefix to the regex literals in `compiler_gym/third_party/inst2vec/inst2vec_preprocess.py`.


compiler_gym/third_party/inst2vec/inst2vec_preprocess.py
```diff
@@ -339,16 +339,16 @@ def preprocess(data):
 # XFG-transforming (inline and abstract statements)
 ########################################################################################################################
 # Helper regexs for structure type inlining
-vector_type = "<\d+ x " + rgx.first_class_type + ">"
-array_type = "\[\d+ x " + rgx.first_class_type + "\]"
-array_of_array_type = "\[\d+ x " + "\[\d+ x " + rgx.first_class_type + "\]" + "\]"
+vector_type = r"<\d+ x " + rgx.first_class_type + r">"
+array_type = r"\[\d+ x " + rgx.first_class_type + r"\]"
+array_of_array_type = r"\[\d+ x " + r"\[\d+ x " + rgx.first_class_type + r"\]" + r"\]"
```
hughleat, line 344: Consider rf-strings?
```python
array_of_array_type = rf"\[\d+ x \[\d+ x {rgx.first_class_type}\]\]"
```

compiler_gym/third_party/inst2vec/rgx_utils.py
```diff
@@ -29,58 +29,58 @@
...
+    s = r"(?:"
     if len(to_add) > 0:
-        s += possibilities[0] + to_add + " "
+        s += possibilities[0] + to_add + r" "
     else:
         s += possibilities[0]
     for i in range(len(possibilities) - 1):
         if len(to_add) > 0:
-            s += "|" + possibilities[i + 1] + to_add + " "
+            s += r"|" + possibilities[i + 1] + to_add + r" "
```
hughleat, line 47: Not that it's wrong, but... why? I can see it when you have backslashes, but there are lots of changes here with no backslashes. Is it sensible to use additional abstractions which aren't necessary?
IMO, only use `r`-strings where necessary. Use `f`-strings or `rf`-strings a lot more.

hughleat (approved):
LGTM

compiler_gym/third_party/inst2vec/rgx_utils.py
ChrisCummins: I agree. Its an automatic refactoring of third party code. Rewriting the regexes to `rf`-strings would be nice, but I don't know how to do it automatically

PR #606, opened 2022-03-05
Change CompilerEnv.step to accept a single action

sogartar:


ChrisCummins (changes requested):
Hi @sogartar,

I agree that CompilerEnv.step should accept a single action, but there's a couple of things we need to do before we make this change:

1. We need to add a deprecation warning on `env.step([1, 2, 3])`, and cut a release of CompilerGym with this warning so that users get advance notice that their code will break.
1. We need to add a new way for specifying a list of actions to run in a single step. The benefit of `env.step(actions)` over `for a in actions: env.step(a)` is that the former can be executed in a single RPC invocation, significantly reducing the overhead of round trips to the backend, and removing the need to calculate observation/rewards for each individual step. We measured speedups of ~3x on typical LLVM workloads using this (more details [here](https://arxiv.org/pdf/2109.08267.pdf)).

I have a draft PR for (2) already work-in-progress. If you're happy for me to, I can pull in some of your other fixes from this commit and send you the PR when ready.

I made a tracking issue for this, #610.

Cheers,
Chris

sogartar:
I will add the multistep to this 

> I have a draft PR for (2) already work-in-progress. If you're happy for me to, I can pull in some of your other fixes from this commit and send you the PR when ready.

That is OK. In this PR I change a lot of tests to not use mltistep actions. These pieces would have to be reverted.

ChrisCummins:
Thanks! I cherry-picked your commit into #606 and selectively reverted some of the test refactors. It's still WIP, need to figure out the best way for wrappers to interact with multiple different step() implementations.

Cheers,
Chris

ChrisCummins:
Superseded by #627.

PR #565, opened 2022-02-06
Rename id attribute for rewards to "name"

parthchadha:
This PR renames the `id` attribute in `Reward` class to `name` to be consistent with other space classes. 

Fixes #381

Issue #381, opened 2021-09-08
Rename Reward.id to Reward.name


ChrisCummins:
## 🚀 Feature

The `Reward` class is unusual in that its name attribute is called `id`. We should rename this to `name` to be consistent with all of the other space classes.

## Motivation

To be consistent.

PuneetDabral:
sir can i do my first contribution here ...but i need your support

ChrisCummins:
Hi @PuneetDabral, great! I can help you.

The first thing to do would be to follow the [install instructions](https://github.com/facebookresearch/CompilerGym/blob/development/INSTALL.md#building-from-source) to build CompilerGym from source on your local machine.

The Reward class can be found in `compiler_gym/spaces/reward.py`. The job of this task is to rename the [Reward.id](https://github.com/facebookresearch/CompilerGym/blob/5354dddb18e721b55335434524bd2fa72fb3ca2e/compiler_gym/spaces/reward.py#L84) attribute.

It might take a while to check through the source code for occurrences of the `id` name. Be sure to run the unit tests, they will help.

Good luck! Let me know if you have any questions.

Cheers,
Chris

PuneetDabral:
sir i changed self.id=name now what i can do?? can you please guide me

PR #612, opened 2022-03-07
Add backwards compatibility for Reward.id.

ChrisCummins:
We don't want the change introduced in #565 to immediately break
existing user code. Instead, add a workaround that emits a deprecation
warning so that users can update their code upon the next release.

Issue #381.


mostafaelhoushi (approved):
Cool! Wondering why the Cmake CI job is failing :(

UPDATE: it is not Cmake failing, but the pull job. Hopefully that is simple

Issue #595, opened 2022-03-03
Speedup CMake Configuration


mostafaelhoushi:
## 🚀 Feature

Speedup CMake Configuration Step

## Motivation

Everytime we update a CMakeLists.txt file, cmake build/configuration takes a really long time. Probably, it has to re-run the entire configuration, and that means rebuilding LLVM and a bunch of stuff from source. 

## Pitch

N/A

## Alternatives

N/A 

## Additional context

Some projects have rolled their own ad-hoc way to prevent rebuilds, like https://github.com/facebookresearch/CompilerGym/blob/development/external/protobuf/build_protobuf.cmake#L8-L9

so we could use a similar approach

Cc @ChrisCummins

ChrisCummins:
Thanks for filing this.

I think the goal here will be to either:

1. build less stuff from source, by using system-wide deps or depending on binary releases
2. figure out how to smartly prevent rebuilding deps when the config changes

The (1) option would be useful for the CI as well, as atm the `build-linux-cmake` step takes **5 hours**, which is a productivity drain. The real time-sink is building LLVM from source, so the fix for this might be related to #568.

Cheers,
Chris

mostafaelhoushi:
Also I am wondering, if a quick temporary speedup of the LLVM build would be to add `-j32` to its build command.

ChrisCummins:
> Also I am wondering, if a quick temporary speedup of the LLVM build would be to add `-j32` to its build command.

As far as I understand, the LLVM build is hardcoded to use Ninja, so should already be achieving high levels of parallelism:
https://github.com/facebookresearch/CompilerGym/blob/development/external/llvm/CMakeLists.txt#L43-L44

PR #618, opened 2022-03-09
Build system tweaks patchset

ChrisCummins:
This pull request makes two important changes to the build:

- Fixes a bunch of errors that were introduced in #620 (not sure what happened there), un-breaking the build.
- Adds support for building CompilerGym against a pre-compiled LLVM. This reduces CI build time by **35 minutes**, and reduces the size of the build directory by **55%** (11 GB to 4.9 GB) (#595).

And a bunch of smaller tweaks to the CMake build:
- Switches the CI cmake job to use a pre-compiled LLVM.
- Updates the version of absl and benchmark to work around some build errors on newer linux releases.
- Updates ProGraML so that it uses the same version of absl.
- Turns off the redundant conditional logic for building on Darwin. There is no Linux/macOS-dependent logic in the build system (except for janky linker workarounds needed by bazel).
- Automatically enables ccache in the CMake config when it is available.
- Automatically uses the lld linker when it is available.
- Forces the Ninja generator.
- Makes a number of small formatting tweaks for readability.

---

My eventual goal is to add support for CMake on macOS (#618), but I am hitting a problem with the protobuf configuration step (`autoreconf`) in `external/protobuf/build_protobuf.cmake`:

```
checking for g++ options needed to detect all undeclared functions... cannot detect
configure: error: in `/Users/cummins/src/CompilerGym/build/external/protobuf/protobuf/src/protobuf':
configure: error: cannot make g++ report undeclared builtins
See `config.log' for more details
CMake Error at /Users/cummins/src/CompilerGym/external/protobuf/build_protobuf.cmake:50 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 1"
```

For some reason, the configure script doesn't like the detected C/C++ compiler. I'll pick this up at a later date.

### DIFF
--- a/.github/actions/install-cmake-build-dependencies/action.yaml
+++ b/.github/actions/install-cmake-build-dependencies/action.yaml
@@ -9,8 +9,7 @@ runs:
         - name: Install CMake dependencies
           run: |
               if [ "$(uname)" = "Darwin" ]; then
-                  echo "CMake build for Darwin is unimplemented."
-                  exit 1
+                  brew install bazelisk ccache cmake zlib
               else
                   # Compiler
                   sudo apt-get install clang++-9 lld-9 tar bzip2 ninja-build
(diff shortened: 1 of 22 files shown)


ChrisCummins:
@sogartar could you take a quick look at this?

Cheers,
Chris

ChrisCummins:
Merging now as this fixes the `development` build.

Cheers,
Chris

PR #633, opened 2022-03-19
Refactor out Env interface from CompilerEnv

sogartar:
The goal of this change is to have a clean separation between interface and implementation. It also allows for new environment implementations with approaches different from `CompilerEnv`.


ChrisCummins (commented):
I'm still not 100% sure what the use case is for splitting interface / implementation like this, but I guess there's no harm in it.

However, I think the naming needs rethinking. `Env` implies that it isn't specific to compilers, which I believe it is, so I think the two choices are either:

1. Pick a name that makes it clear it is describing the compiler env, like `CompilerEnvInterface` or `CompilerEnvBase`.
2. Call it `CompilerEnv` and rename the current `CompilerEnv` to make it clear that it is an implementation of CompilerEnv, like `ClientServiceCompilerEnv` or even just `CompilerEnvImplementation`.

I think option 2 makes more sense to me. That way, all of the type hints for `CompilerEnv` now refer to the abstract interface. This may need updating code that refers to `CompilerEnv` and handles implementation details like interacting with `env.service`.

What do you think?

Cheers,
Chris

sogartar:
> I'm still not 100% sure what the use case is for splitting interface / implementation like this, but I guess there's no harm in it.

The idea is to follow the [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) and [open–closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). Users that write algorithms against this library would know to use only this interface and it will be stable upon changing the implementation. It may not be planed currently to have a different compiler service, but this leaves the door open. It makes clear to someone else who wants to make an environment what is the required interface to work with the algorithms.

> However, I think the naming needs rethinking.

My intention was to not introduce a breaking change and since `Env` is in the `compiler_gym` package it should be clear that it is about compilers. Renaming the existing `CompilerEnv` class would be a breaking change for someone who extends it.
Aside form the breaking change the best option would be `CompilerEnv` for the interface and `ClientServiceCompilerEnv` for the implementation. Would you like me to use these names?

ChrisCummins:
Thanks @sogartar!

> It may not be planed currently to have a different compiler service, but this leaves the door open.

Yeah, I don't know if there's demand for this. But, as I said, I don't think it'll hurt, and it does seem like better practice to clearly delineate interface/implementation 🙂 

> Renaming the existing `CompilerEnv` class would be a breaking change for someone who extends it. Aside form the breaking change the best option would be `CompilerEnv` for the interface and `ClientServiceCompilerEnv` for the implementation. Would you like me to use these names?

Yes, I think this is worth a breaking change. Make `CompilerEnv` in `compiler_env.py` the interface, and rename the current CompilerEnv to a name that describes the implementation mechanism. I think `ClientServiceCompilerEnv` in `client_service_compiler_env.py` is a good choice for this, though I'm open to suggestions.

Cheers,
Chris

sogartar:
@ChrisCummins, I made the renaming. Can you let me know what else should be included or removed in the interface? I don't think I can make the call. I am not sure that the `observation_space_spec` property should be a part or the benchmark or datasets related stuff.
I am also not sure of all the places were the interface should be used or the service. It really depends what should be in the interface.
The basic structure should be:
```
CompilerEnv <- ClientServiceCompilerEnv <- LllvmEnv, GccEnv, etc.
      ^
      |
Algorithms
```
Ideally, general algorithms should depend only on CompilerEnv.
I feel like the service client may be another important concept separate from the CompilerEnv interface and the gRPC specifics. It should define common service functionality, but that is a discussion for another time.

benchmarks/bench_test.py
```diff
@@ -21,7 +21,7 @@
 import pytest
 
 import examples.example_compiler_gym_service as dummy
-from compiler_gym.envs import CompilerEnv, LlvmEnv, llvm
+from compiler_gym.envs import ClientServiceCompilerEnv, LlvmEnv, llvm
```
ChrisCummins, line 24: nn

compiler_gym/__init__.py
```diff
@@ -34,7 +34,7 @@
     CompilerEnvStateReader,
     CompilerEnvStateWriter,
 )
-from compiler_gym.envs import COMPILER_GYM_ENVS, CompilerEnv
+from compiler_gym.envs import COMPILER_GYM_ENVS, ClientServiceCompilerEnv
```
ChrisCummins, line 37: nn

compiler_gym/envs/client_service_compiler_env.py
```diff
@@ -0,0 +1,1372 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
+#
+# This source code is licensed under the MIT license found in the
+# LICENSE file in the root directory of this source tree.
+
```
ChrisCummins: Needs a module docstring saying that this module defines an implementation of CompilerEnv interface using client/server approach

compiler_gym/envs/client_service_compiler_env.py
```diff
@@ -0,0 +1,1372 @@
...
+    try:
+        return service(service.stub.Step, request)
+    except FileNotFoundError as e:
+        if str(e).startswith("Session not found"):
+            raise SessionNotFound(str(e))
+        raise
+
+
+class ClientServiceCompilerEnv(CompilerEnv):
+    """An OpenAI gym environment for compiler optimizations.
```
ChrisCummins: This docstring should be on the `CompilerEnv` base class

compiler_gym/envs/client_service_compiler_env.py
```diff
@@ -0,0 +1,1372 @@
...
+
+        :param service_connection: An existing compiler gym service connection
+            to use.
+
+        :raises FileNotFoundError: If service is a path to a file that is not
+            found.
+
+        :raises TimeoutError: If the compiler service fails to initialize within
+            the parameters provided in :code:`connection_settings`.
+        """
```
ChrisCummins: Does this need to invoke gym.Env superclass constructor?

compiler_gym/envs/compiler_env.py
```diff
@@ -3,561 +3,30 @@
...
-        See also :meth:`commandline()
-        <compiler_gym.envs.CompilerEnv.commandline>`.
-
-        Calling this method on a :class:`CompilerEnv
-        <compiler_gym.envs.CompilerEnv>` instance raises
-        :code:`NotImplementedError`.
-
-        :return: A list of actions.
-        """
+    @observation_space_spec.setter
```
ChrisCummins, line 22: I'm not sure we need abstract property setters

compiler_gym/envs/compiler_env.py
```diff
@@ -1051,38 +89,8 @@ def step(  # pylint: disable=arguments-differ
...
-            warnings.warn(
-                "Argument `rewards` of CompilerEnv.step has been renamed "
-                "`reward_spaces`. Please update your code",
-                category=DeprecationWarning,
-            )
-            reward_spaces = rewards
-        return self.multistep([action], observation_spaces, reward_spaces)
+        raise NotImplementedError("abstract method")
 
     def multistep(
```
ChrisCummins, line 335:
```suggestion
    @abstractmethod
    def multistep(
```

ChrisCummins:
> @ChrisCummins, I made the renaming. Can you let me know what else should be included or removed in the interface? I don't think I can make the call. I am not sure that the `observation_space_spec` property should be a part or the benchmark or datasets related stuff. I am also not sure of all the places were the interface should be used or the service. It really depends what should be in the interface. The basic structure should be:
> 
> ```
> CompilerEnv <- ClientServiceCompilerEnv <- LllvmEnv, GccEnv, etc.
>       ^
>       |
> Algorithms
> ```

Thanks for the update @sogartar. IMO _everything_ that isn't solely relevant to the client/server design should be part of the `CompilerEnv` interface. Basically, extend "Algorithms" in your above diagram to "All user code".

This would make `ClientServiceCompilerEnv` an implementation detail that can be ignored by most users, and the only places where `CompilerEnv` needs updating to `ClientServiceCompilerEnv` is code which directly interacts with `env.service` or `env.raw_step()` (a handful of unit tests). The current `CompilerEnv` interface with only step/multistep/fork would require massive refactors to replace `CompilerEnv` with `ClientServiceCompilerEnv`, which seems like it would be going against the goal of separating interface from implementation, as 99.9% of the time no one cares if its client/server. Sorry if I wasn't clear on this, and I hope that your rename was an auto find+replace so it isn't a hassle to revert 😄 

> Ideally, general algorithms should depend only on CompilerEnv. I feel like the service client may be another important concept separate from the CompilerEnv interface and the gRPC specifics. It should define common service functionality, but that is a discussion for another time.

I wondered about whether it should be `GrpcClientServiceCompilerEnv`, as you could in theory have a different client/server architecture that doesn't use subprocesses + gRPC, but I think thats a bridge to cross when needed. It would require larger refactor to abstract the idea of a "service" from gRPC + protos.

One that could be worth doing now to highlight that `ClientServiceCompilerEnv` is an implementation detail would be to move the file from `compiler_gym/envs/client_service_compiler_env.py` into `compiler_gym/service/client_service_compiler_env.py`. That would then group all of the code relating to protos/grpc in `compiler_gym/service`.

Another question is whether `CompilerEnv` _needs_ to be abstract. E.g. take a property like:

```py
    @property
    @abstractmethod
    def episode_walltime(self) -> float:
        raise NotImplementedError("abstract method")
```

this seems like it would be straightforward enough bit of housekeeping for `CompilerEnv`  to provide as it clearly don't depend on the implementation details of the environment. But that's always something that can be done later and isn't blocking for this PR.

Cheers,
Chris

sogartar:
Some of the details are not properly isolated. For example `ObservationSpaceSpec` and `Benchmark` has protobuf related stuff in their interfaces. I can add to `CompilerEnv` the related functionality with the knowledge that this would break the isolation and then at a later stage fix them as well.

ChrisCummins:
Yeah, I think they are candidates for future refactoring, but we unfortunately can't get away from dropping them completely from the `CompilerEnv` interface without having to update a lot of code.

Cheers,
Chris

compiler_gym/envs/client_service_compiler_env.py
sogartar: done

compiler_gym/envs/client_service_compiler_env.py
sogartar: I moved a part of it to `CompilerEnv`.

compiler_gym/envs/client_service_compiler_env.py
sogartar: Neither `CompilerEnv` or `gym.Env` have `__init__(...)`.

compiler_gym/envs/compiler_env.py
sogartar: done

sogartar:
@ChrisCummins, I think this PR is ready for another code-review. I did what you suggested. One thing to note is that moving `ClientServiceCompilerEnv` to `service` caused a circular dependency that I broke by not making `ClientServiceCompilerEnv` a dependency of the `service` package. I think this will auto-resolve if benchmark, dataset, etc. do not depend on service.

sogartar:
I am not sure why this failure https://github.com/facebookresearch/CompilerGym/runs/5669835818?check_suite_focus=true#step:8:213 started appearing now. It should have triggered before also, since `gym.Wrapper` redirects each attribute to the underlying env. What I did is to add the property explicitly in `CompilerEnvWrapper` for better stack tracing.

ChrisCummins:
> I am not sure why this failure https://github.com/facebookresearch/CompilerGym/runs/5669835818?check_suite_focus=true#step:8:213 started appearing now. It should have triggered before also, since `gym.Wrapper` redirects each attribute to the underlying env. What I did is to add the property explicitly in `CompilerEnvWrapper` for better stack tracing.

Is the problem the UserWarning?

compiler_gym/envs/compiler_env.py
```diff
@@ -3,85 +3,21 @@
...
-    try:
-        return service(service.stub.Step, request)
-    except FileNotFoundError as e:
-        if str(e).startswith("Session not found"):
-            raise SessionNotFound(str(e))
-        raise
-
-
-class CompilerEnv(gym.Env):
+class CompilerEnv(gym.Env, ABC):
```
ChrisCummins, line 21: I think there are still a few properties missing from this interface: `version`, `compiler_version` (possibly others)

compiler_gym/envs/compiler_env.py
```diff
@@ -1109,148 +295,47 @@ def multistep(
...
-
-        CompilerEnv instances support two render modes: "human", which prints
-        the current environment state to the terminal and return nothing; and
-        "ansi", which returns a string representation of the current environment
-        state.
-
-        :param mode: The render mode to use.
-        :raises TypeError: If a default observation space is not set, or if the
-            requested render mode does not exist.
         """
```
ChrisCummins: I think this should be on `CompilerEnv`. Is the reason it isn't because it is also on `gym.Env`?

compiler_gym/random_replay.py
```diff
@@ -8,7 +8,7 @@
 
 from deprecated import deprecated
 
-from compiler_gym.envs.compiler_env import CompilerEnv
```
ChrisCummins: IRRC, this is actually useful. It means that `//compiler_gym:random_replay` need only depend on `//compiler_gym/envs:compiler_env` rather than `//compiler_gym`

compiler_gym/service/BUILD
```diff
@@ -12,6 +12,8 @@ py_library(
     deps = [
         ":compilation_session",
         ":connection",
+        # TODO: add this after circular dependencies are resolved
+        # ":client_service_compiler_env",
```
ChrisCummins: I wonder if the easies thing here would be to merge `:service` and `:client_service_compiler_env`, so targets need only depend on `//compiler_gym/service`?

ChrisCummins: (same for cmake)

compiler_gym/service/client_service_compiler_env.py
```diff
@@ -0,0 +1,1291 @@
...
+    :ivar reward: A view of the available reward spaces that permits on-demand
+        computation of rewards.
+
+    :vartype reward: compiler_gym.views.RewardView
+
+    :ivar episode_reward: If :func:`ClientServiceCompilerEnv.reward_space
+        <compiler_gym.envs.CompilerGym.reward_space>` is set, this value is the
+        sum of all rewards for the current episode.
+
+    :vartype episode_reward: float
```
ChrisCummins: I think each of these `ivar` comments should a docstring on the corresponding abstract property in `CompilerEnv`.

examples/example_compiler_gym_service/__init__.py
```diff
@@ -83,7 +83,7 @@ def benchmark_from_parsed_uri(self, uri: BenchmarkUri) -> Benchmark:
 # the example-v0 environment will be available to gym.make(...).
 register(
     id="example-cc-v0",
-    entry_point="compiler_gym.envs:CompilerEnv",
+    entry_point="compiler_gym.service.client_service_compiler_env:ClientServiceCompilerEnv",
```
ChrisCummins, line 86: It would be nice to add `ClientServiceCompilerEnv` to the `__all__` list in `compiler_gym/service/__init__.py`, so this entry point would be `compiler_gym.service:ClientServiceCopmilerEnv`.

ChrisCummins: and for the others in this file

examples/loop_optimizations_service/__init__.py
```diff
@@ -137,7 +137,7 @@ def benchmark_from_parsed_uri(self, uri: BenchmarkUri) -> Benchmark:
 
 register(
     id="loops-opt-py-v0",
-    entry_point="compiler_gym.envs:CompilerEnv",
+    entry_point="compiler_gym.service.client_service_compiler_env:ClientServiceCompilerEnv",
```
ChrisCummins, line 140: Suggest `compiler_gym.service:ClientServiceCompilerEnv`

ChrisCummins (commented):
Hi @sogartar, this is coming along great! I've come around to like the separation of interface / gRPC implementation and am already thinking of environments I'd like to implement that aren't based on this client / service design 🙂 

Last few nitpicky requests left inline.

Also, I think there's a handful of files where the class name has been updated where it doesn't need to be. Please revert them (unless I'm wrong!):

```
benchmarks/bench_test.py
compiler_gym/bin/service.py
compiler_gym/envs/llvm/llvm_benchmark.py
compiler_gym/random_replay.py
compiler_gym/util/debug_util.py
examples/llvm_autotuning/autotuners/opentuner_.py
examples/llvm_rl/model/inference_result.py
leaderboard/llvm_instcount/random_search/README.md
```

Cheers,
Chris

ChrisCummins:
Oh, and the new `ClientServiceCompilerEnv` needs to be added to the documentation. I'd suggest this page https://compilergym.com/compiler_gym/service.html

examples/example_compiler_gym_service/__init__.py
sogartar: This would cause circular import. I tried it already. I am not sure how to resolve the issue. The problem is that stuff in `dataset` and `view` depend on service and by doing this service would depend on `dataset` and `view`. The correct solution would be to also split `dataset` and `view` into interface and implementation and have ClientServiceCopmilerEnv depend on the interfaces.

sogartar:
> > I am not sure why this failure https://github.com/facebookresearch/CompilerGym/runs/5669835818?check_suite_focus=true#step:8:213 started appearing now. It should have triggered before also, since `gym.Wrapper` redirects each attribute to the underlying env. What I did is to add the property explicitly in `CompilerEnvWrapper` for better stack tracing.
> 
> Is the problem the UserWarning?

It seems it is. The test should probably expect the warning.

compiler_gym/service/BUILD
sogartar: This will cause circular dependency. I have more elaborate explanation in another comment.

examples/example_compiler_gym_service/__init__.py
sogartar: If `ClientServiceCopmilerEnv` needs the details, they would be implementations related to the gRPC service, so they would also be able to move to `service`.

compiler_gym/envs/compiler_env.py
sogartar: I added them and a few more.

compiler_gym/envs/compiler_env.py
sogartar: The reason was that it is a part of `gym.Env`, but I added it in `CompilerEnv` as well. It has a type hint and the additional documentation.

compiler_gym/random_replay.py
sogartar: Done

compiler_gym/service/client_service_compiler_env.py
sogartar: done

examples/loop_optimizations_service/__init__.py
sogartar: See my other comments about circular dependencies.

sogartar:
> Oh, and the new `ClientServiceCompilerEnv` needs to be added to the documentation. I'd suggest this page https://compilergym.com/compiler_gym/service.html

I added it there, but technically `ClientServiceCompilerEnv` is not a part of service due to the circular dependency.

sogartar:
> Also, I think there's a handful of files where the class name has been updated where it doesn't need to be. Please revert them (unless I'm wrong!):
> 
> ```
> benchmarks/bench_test.py
> compiler_gym/bin/service.py
> compiler_gym/envs/llvm/llvm_benchmark.py
> compiler_gym/random_replay.py
> compiler_gym/util/debug_util.py
> examples/llvm_autotuning/autotuners/opentuner_.py
> examples/llvm_rl/model/inference_result.py
> leaderboard/llvm_instcount/random_search/README.md
> ```

I renamed most of them.
I think these need `ClientServiceCompilerEnv`:
```
benchmarks/bench_test.py
examples/llvm_autotuning/autotuners/opentuner_.py
```
They use stuff like `send_param` and `service.connection`

sogartar:
The PR is ready of another review.

sogartar:
@ChrisCummins, do you have an idea why the [pre-commit checks](https://github.com/facebookresearch/CompilerGym/runs/5699842822?check_suite_focus=true) are failing?

sogartar:
One thing to mention is that the `ObservationWrapper` and `RewardWrapper` needed to have their conversion methods renamed because they conflict with methods from the `CompilerEnv` base. These are `observation(...)` and `reward(...)`.

ChrisCummins:
> @ChrisCummins, do you have an idea why the [pre-commit checks](https://github.com/facebookresearch/CompilerGym/runs/5699842822?check_suite_focus=true) are failing?

No, sorry. Looks like its a general error, not specific to these changes. Does `development` branch fail?

> I think these need ClientServiceCompilerEnv:
> ...
> They use stuff like send_param and service.connection

Good catch!

compiler_gym/service/BUILD
ChrisCummins: Okay, in that case could you update the TODO note to reference this?

```
TODO(github.com/facebookresearch/CompilerGym/pull/633): ...
```

examples/example_compiler_gym_service/__init__.py
ChrisCummins: Ah I see, thanks for the context. Okay, let's keep it the way you have done. Long term, I think it makes to break `dataset` and `view` dependencies on `service`. Maybe through an equivalent refactor between the dataset/view "interface" and gRPC implementation?

tests/wrappers/core_wrappers_test.py
```diff
@@ -18,9 +19,30 @@
 pytest_plugins = ["tests.pytest_plugins.llvm"]
 
 
+class ObservationDummyWrapper(ObservationWrapper):
```
ChrisCummins: Why `ObservationDummyWrapper` and not `ObservationWrapper` (using an `import .. as ..` to avoid shadowing the `ObservationWrapper` in gym module)

tests/wrappers/core_wrappers_test.py
```diff
@@ -255,15 +284,13 @@ def observation(self, observation):
 
 
 def test_wrapped_observation_missing_definition(env: LlvmEnv):
-
-    env = ObservationWrapper(env)
-    with pytest.raises(NotImplementedError):
-        env.reset()
+    with pytest.raises(TypeError):
+        env = ObservationWrapper(env)
```
ChrisCummins: Hmm, why is it now a `TypeError`? Could you add a `match="..."` argument to match a bit of the description?

ChrisCummins (commented):
Looking good. There are a couple of CI failures that need addressing prior to merge.

I think the `observation()` to `convert_observation()` rename makes sense

Cheers,
Chris

sogartar:
> Ah I see, thanks for the context. Okay, let's keep it the way you have done. Long term, I think it makes to break dataset and view dependencies on service. Maybe through an equivalent refactor between the dataset/view "interface" and gRPC implementation?

This approach should work.

tests/wrappers/core_wrappers_test.py
sogartar: I made the class `ObservationWrapper` an abstract base class, so if you miss to implement some abstract method it raises a `TypeError`. There is no match because I am not confident that the exception message string that `ABC` uses will be stable.

tests/wrappers/core_wrappers_test.py
sogartar: Done.

sogartar:
> Okay, in that case could you update the TODO note to reference this?
> ```
> TODO(github.com/facebookresearch/CompilerGym/pull/633): ...
> ```

Done.

sogartar:
> > @ChrisCummins, do you have an idea why the [pre-commit checks](https://github.com/facebookresearch/CompilerGym/runs/5699842822?check_suite_focus=true) are failing?
> 
> No, sorry. Looks like its a general error, not specific to these changes. Does `development` branch fail?

It seems it does 6cf49267b7b5338de8c070dabb7ea595a13715a9.

sogartar:
@ChrisCummins, could you do another review round?

ChrisCummins:
LGTM, thanks so much @sogartar!

PR #641, opened 2022-03-30
Make action spaces and action service message conversion configurable in ClientServiceCompilerEnv

sogartar:
This is another perquisite for the MLIR env. It enables most actions that are proper in the sense of OpenAI gym to be used in a gRPC compiler env.

One thing to note is that with this change the conversion of messages from/to actions, observations and spaces is a bit spread out. For example the `ObservationSpaceSpec` does the observation conversion. It is still static and does not allow customization at that level. You would have to define a new `ObservationView` like class and override `_observation_view_type` form `ClientServiceCompilerEnv`.

### DIFF
--- a/compiler_gym/service/client_service_compiler_env.py
+++ b/compiler_gym/service/client_service_compiler_env.py
@@ -12,7 +12,7 @@ from copy import deepcopy
 from math import isclose
 from pathlib import Path
 from time import time
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 import numpy as np
 from deprecated.sphinx import deprecated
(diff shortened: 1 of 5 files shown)


compiler_gym/service/client_service_compiler_env.py
```diff
@@ -82,6 +82,27 @@ def _wrapped_step(
         raise
 
 
+class ServiceMessageConverters:
```
ChrisCummins, line 85: This class needs a docstring, and an entry in the restructured text docs

compiler_gym/service/client_service_compiler_env.py
```diff
@@ -156,6 +178,8 @@ def __init__(
             passed to :meth:`env.observation.add_derived_space()
             <compiler_gym.views.observation.Observation.add_derived_space>`.
 
+        :param service_message_converters: custom converters for action spaces and actions.
```
ChrisCummins:
```suggestion
        :param service_message_converters: Custom converters for action spaces and actions.
```
nitpick

ChrisCummins (approved):
Thanks @sogartar, I like this. Just needs a bit of documentation then good to go.

Cheers,
Chris

compiler_gym/service/client_service_compiler_env.py
sogartar: Done.

compiler_gym/service/client_service_compiler_env.py
sogartar: Done.

ChrisCummins (approved):
LGTM, thanks @sogartar!

Issue #526, opened 2021-12-21
Nested spaces in gRPC


sogartar:
## ❓ Questions and Help

It looks like that the proto definitions of [action space](https://github.com/facebookresearch/CompilerGym/blob/398b39f4cda06f43ee119c9b488aaf7b45266f79/compiler_gym/service/proto/compiler_gym_service.proto#L129), [action](https://github.com/facebookresearch/CompilerGym/blob/398b39f4cda06f43ee119c9b488aaf7b45266f79/compiler_gym/service/proto/compiler_gym_service.proto#L139), [observation space](https://github.com/facebookresearch/CompilerGym/blob/398b39f4cda06f43ee119c9b488aaf7b45266f79/compiler_gym/service/proto/compiler_gym_service.proto#L230) and [observation](https://github.com/facebookresearch/CompilerGym/blob/398b39f4cda06f43ee119c9b488aaf7b45266f79/compiler_gym/service/proto/compiler_gym_service.proto#L183) do not support nesting in.
On the other hand OpenAI Gym's interface has [Dict](https://github.com/openai/gym/blob/616b07115871588a5b63fa470d4f554d114d3e25/gym/spaces/dict.py#L7) and by extension CompilerGym has also [Dict](https://github.com/facebookresearch/CompilerGym/blob/398b39f4cda06f43ee119c9b488aaf7b45266f79/compiler_gym/spaces/dict.py#L12). That makes deep hierarchies possible there.

Does It make sense to extend the gRPC interface to support this type of structure?

## Additional Context

This is not a very pressing issue as currently the structure I want to represent is simple enough, that I can use flat names like `a.b`
and `a.c`.
In the future there may be the need for more complex structures.

ChrisCummins:
Hi @sogartar,

> Does It make sense to extend the gRPC interface to support this type of structure?

Interesting suggestion! Possibly 🙂 I can see the value of nested action and observation spaces, but I don't know of any immediate use cases so I haven't prioritized it. The closest we have in the current version would be something like the JSON data structure returned by some of the LLVM environment's observations, but in that case I just send it over as raw bytes data and deserialize on the frontend.

Cheers,
Chris

sogartar:
Currently the structure for the action spaces is

```proto
message ActionSpace {
  string name = 1;
  repeated ChoiceSpace choice = 2;
  bool named_choices = 3;
}

message ChoiceSpace {
  string name = 1;
  oneof space {
    ...
  }
}
```

[...]

PR #643, opened 2022-03-31
Add dispatching on type_id during action and observation conversion

sogartar:
Please see https://github.com/facebookresearch/CompilerGym/issues/526#issuecomment-1011252741 for more information.

### DIFF
--- a/compiler_gym/service/proto/py_converters.py
+++ b/compiler_gym/service/proto/py_converters.py
@@ -90,6 +90,25 @@ class TypeBasedConverter:
         return self.conversion_map[type(val)](val)
 
 
+class TypeIdDispatchConverter:
+    conversion_map: DictType[str, Callable[[Message], Any]]
+    default_converter: Callable[[Message], Any]
+
+    def __init__(
+        self,
+        default_converter: Callable[[Message], Any],
+        conversion_map: DictType[str, Callable[[Message], Any]] = None,
+    ):
+        self.conversion_map = {} if conversion_map is None else conversion_map
+        self.default_converter = default_converter
+
+    def __call__(self, message: Message) -> Any:
+        if message.HasField("type_id"):
+            return self.conversion_map[message.type_id](message)
+        else:
+            return self.default_converter(message)
+
+
 proto_type_to_dtype_map = {
     BooleanTensor: bool,
     ByteTensor: np.int8,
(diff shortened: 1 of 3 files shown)


compiler_gym/service/proto/py_converters.py
```diff
@@ -90,6 +90,25 @@ def __call__(self, val: Any) -> Any:
         return self.conversion_map[type(val)](val)
 
 
+class TypeIdDispatchConverter:
```
ChrisCummins, line 93: Needs a docstring, with a usage example of how to instantiate it

ChrisCummins (approved):
Nice! Code LGTM, just needs a bit of documentation

compiler_gym/service/proto/py_converters.py
sogartar: I added as an example the existing unit test.
Ultimately, there should be a short guide on how to customize the conversion from/to proto.
I think the current method here and other converter classes is not good enough. They use dispatching based on dict keys, but this is not flexible enough when you have to convert for example any `numbers.Integral`. After an attempt on conversion with an exact type match it should go through a list of fallback converters that attempt the conversion one-by-one and if none succeeds there should be a failure.

compiler_gym/service/proto/py_converters.py
```diff
@@ -91,6 +91,36 @@ def __call__(self, val: Any) -> Any:
 
 
 class TypeIdDispatchConverter:
+    """Dispatches conversion of a <google.protobuf.message.Message>
+    based on the value of its field "type_id".
+
+    If the "type_id" filed is not present the conversion falls back on `default_converter`.
+    Example:
+    ```
```
ChrisCummins: Nitpick: docstrings need to use rst syntax, so this should use `.. code-block:: python` rather than ```.

compiler_gym/service/proto/py_converters.py
ChrisCummins: SGTM. Could you file a documentation issue requesting this?

sogartar: I opened one at #649.

compiler_gym/service/proto/py_converters.py
sogartar: Done.

ChrisCummins (approved):
LGTM!

PR #644, opened 2022-03-31
Include name in space comparison

sogartar:
I am not entirely sure if this is the desired behavior.

### DIFF
--- a/compiler_gym/spaces/box.py
+++ b/compiler_gym/spaces/box.py
@@ -38,3 +38,10 @@ class Box(GymBox):
         """
         super().__init__(low=low, high=high, shape=shape, dtype=dtype)
         self.name = name
+
+    def __eq__(self, other) -> bool:
+        return (
+            isinstance(self, other.__class__)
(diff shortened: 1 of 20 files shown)


ChrisCummins (commented):
This seems reasonable. I can't think of a why you wouldn't want to compare on name.

Could you add a positive and negative unit test for each of the `__eq__` operators?

Cheers,
Chris

sogartar:
> Could you add a positive and negative unit test for each of the `__eq__` operators?

Done.

ChrisCummins:
Thanks a lot!

The `id` argument for `Reward` is deprecated. If you remove them that should fix the CI, then this looks ready to merge.

Cheers,
Chris

sogartar:
> The `id` argument for `Reward` is deprecated. If you remove them that should fix the CI, then this looks ready to merge.

I removed it and fixed the other tests.

ChrisCummins (approved):
LGTM, thanks @sogartar!

PR #665, opened 2022-04-27
[examples] Fix path to clang in example services.

ChrisCummins:
Use the version of clang that is bundled with CompilerGym, since the
user may not have a compatible version of clang in their $PATH.

### DIFF
--- a/examples/example_unrolling_service/service_py/example_service.py
+++ b/examples/example_unrolling_service/service_py/example_service.py
@@ -245,7 +245,7 @@ class UnrollingCompilationSession(CompilationSession):
             # build object file to binary
             run_command(
                 [
-                    "clang",
+                    self._clang,
                     self._obj_path,
                     "-O3",
                     "-o",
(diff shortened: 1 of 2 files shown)


mostafaelhoushi (approved):
LGTM. Did you make sure that the related tests work? They are still not in CI.

Maybe try:
```
bazel run //examples/example_unrolling_service:example
```

and
```
bazel run //examples/loop_optimizations_service:example
```

to ensure they are working

ChrisCummins:
I noticed the failure on my local dev machine that doesn't have `clang`. With this patch:

```
$ bazel run //examples/example_unrolling_service:example

INFO: Build completed successfully, 1287 total actions
observation:  [12  0 15]

observation:  [18  0 21]
reward:  0.019138755980861243
done:  False
info:  {'action_had_no_effect': False, 'new_action_space': False}

observation:  [18  0 21]
reward:  0.00797448165869214
done:  False
info:  {'action_had_no_effect': True, 'new_action_space': False}
```

Does that look right to you?

Cheers,
Chris

mostafaelhoushi:
> Does that look right to you?
Yes, this is perfect!

PR #666, opened 2022-04-27
Some small cmake tweaks

ChrisCummins:
Some small tweaks to attempt to reduce the number of flags needed to pass to cmake for most use cases:

- `Python3_FIND_VIRTUALENV=FIRST` is the default value, so don't specify it in install instructions.
- Define `CMAKE_BUILD_WITH_INSTALL_RPATH` in `CMakeLists.txt` rather than from command line.
- Don't override command line flags to set ccache / lld paths.


sogartar:
I don't think that the CompilerGym Python package utilizes rpath. It should also be installable anywhere. We don't install through CMake's `install` target. `CMAKE_BUILD_WITH_INSTALL_RPATH` will hardcode the rpath based on `CMAKE_INSTALL_PREFIX`.

ChrisCummins:
Is it a good idea to hardcode it as I have done here?

ChrisCummins:
Deferring to you on this @sogartar  as the cmake expert 🙂

sogartar:
I think we don't need `CMAKE_BUILD_WITH_INSTALL_RPATH` because the service executables link statically to their dependencies.

sogartar (approved):

PR #673, opened 2022-05-02
In install instructions source ~/.profile

sogartar:
The CMake installation may be the first to create ~/.local/bin,
so it would not be in PATH without logging in again.
This change makes sure that it is added to PATH in the same shell session.


ChrisCummins:
Is `~/.profile` standard? None of my systems have it. How about this?

```
export PATH=$HOME/.local/bin:$PATH
```

Cheers,
Chris

sogartar:
I think it is pretty standard. Whether it will add `~/.local/bin` to PATH is another matter. On Ubuntu `~/.profile` adds `~/.local/bin` to PATH only if exists.

sogartar:
I changed it to what you suggested.

ChrisCummins:
Thanks. LGTM

PR #674, opened 2022-05-04
Fix install insturctions and build/ci

sogartar:
CMake: don't force on the user ccache
CMake: don't override variables when setting linker to lld
In install instructions state that CMake build is still experimental
In install instructions don't install cmake through conda (it is isntalled with wget in another place)


INSTALL.md
```diff
@@ -54,7 +54,7 @@ with the required dependencies:
 
     conda create -y -n compiler_gym python=3.8
     conda activate compiler_gym
-    conda install -y -c conda-forge cmake doxygen pandoc patchelf
```
ChrisCummins: I believe this is actually needed for the bazel build (which runs cmake to compile a couple of deps), and the existing CMake instructions are compatible, because it injects the new cmake at `~/.local/bin` in the $PATH before conda's

ChrisCummins (commented):
LGTM. Small suggestion about not removing cmake from build deps, as it is needed for bazel.

What's wrong with forcing ccache?

Cheers,
Chris

INSTALL.md
sogartar: I added it back in.

sogartar:
> What's wrong with forcing ccache?

I don't feel strongly about this, but the user may want to use the cache for something else that is more important to rebuild faster. It has a size limit.
I could put it back in. But it has to be in a way that does not overwrite the  `CMAKE_*_COMPILER_LAUNCHER` variables because the user may have set it to something else.

ChrisCummins:
> I don't feel strongly about this, but the user may want to use the cache for something else that is more important to rebuild faster. It has a size limit.

Ah I see, that makes sense. Let's get rid of it then

ChrisCummins (approved):
PR is marked Draft so lemme know when you want to merge. Otherwise, LGTM.

Cheers,
Chris

sogartar:
I accidentally made it a draft.

PR #675, opened 2022-05-06
CMake: fix custom target dependencies

### LINKED ISSUE #636: Dependencies of custom commands on custom targets in CMake do not trigger a remake
## 🐛 Bug

From the documentation of CMake's [add_custom_command](https://cmake.org/cmake/help/latest/command/add_custom_command.html):
> DEPENDS
> Specify files on which the command depends. Each argument is converted to a dependency as follows:
> 1. If the argument is the name of a target (created by the [add_custom_target()](https://cmake.org/cmake/help/latest/command/add_custom_target.html#command:add_custom_target), [add_executable()](https://cmake.org/cmake/help/latest/command/add_executable.html#command:add_executable), or [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) command) a target-level dependency is created to make sure the target is built before any target using this custom command. Additionally, if the target is an executable or library, a file-level dependency is created to cause the custom command to re-run whenever the target is recompiled.

This is a legacy problem where custom targets are always considered dirty and rebuilt. There is only a guarantee for the order that the custom command would run after the custom target. But the custom command would not rerun if the custom target changes because the custom target is always ran.
We can fix this by using generator expressions. Each custom target must list in a property all its output files. Then the custom command can depend on them with a generator expression.
Generator expressions are also not ideal, but seem like the only option here. There are some issues with some generator expressions where they can not be used in conditional statements because they raise errors. Unfortunately, generator expressions are evaluated eagerly, not conditionally.
I think this must be fixed before switching to CMake.

sogartar:
This fixes dependency issues where a dependency of a custom command
on a custom target guarantees only the order of building and does
not guarantee that the custom command would rerun every time
when the custom target is rebuilt.

The approach to fix this is for custom targets to record in the
OUTPUTS target property their output files. Then the dependent
custom command depends on the contents of the OUTPUTS property
using generator expressions. This creates a file level dependencies
that triggers rebuilds on change.

Fixes #636

### DIFF
--- a/build_tools/cmake/cg_filegroup.cmake
+++ b/build_tools/cmake/cg_filegroup.cmake
@@ -3,10 +3,18 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+include_guard(GLOBAL)
+include(cg_target_outputs)
+
 function(cg_filegroup)
     cmake_parse_arguments(_ARG "PUBLIC" "NAME" "FILES;DEPENDS" ${ARGN})
     rename_bazel_targets(_NAME "${_ARG_NAME}")
     add_custom_target(${_NAME})
+
+    rename_bazel_targets(_DEPS "${_ARG_DEPENDS}")
+    cg_target_outputs(TARGETS ${_DEPS} RESULT _DEPS_OUTPUTS)
+
+    unset(_OUTPUTS)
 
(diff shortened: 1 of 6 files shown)


ChrisCummins:
I don't fully understand the code so I appreciate the detailed PR description. 🙂  Thanks @sogartar!

Cheers,
Chris

PR #676, opened 2022-05-10
Update bug fix issue template

ChrisCummins:
- Remove `conda` from the list of installation sources (we don't have a conda package)
- Port `collect_env.py` from PyTorch repo to this repo.
- Adapt `collect_env.py` for CompilerGym by removing irrelevant CUDA / PyTorch dependencies.

### DIFF
--- a/build_tools/collect_env.py
+++ b/build_tools/collect_env.py
@@ -1,0 +1,559 @@
+# Based on the 'util/collect_env.py' script from PyTorch.
+# <https://github.com/pytorch/pytorch>
+#
+# From PyTorch:
+#
+# Copyright (c) 2016-     Facebook, Inc            (Adam Paszke)
+# Copyright (c) 2014-     Facebook, Inc            (Soumith Chintala)
+# Copyright (c) 2011-2014 Idiap Research Institute (Ronan Collobert)
+# Copyright (c) 2012-2014 Deepmind Technologies    (Koray Kavukcuoglu)
+# Copyright (c) 2011-2012 NEC Laboratories America (Koray Kavukcuoglu)
+# Copyright (c) 2011-2013 NYU                      (Clement Farabet)
+# Copyright (c) 2006-2010 NEC Laboratories America (Ronan Collobert, Leon Bottou, Iain Melvin, Jason Weston)
+# Copyright (c) 2006      Idiap Research Institute (Samy Bengio)
+# Copyright (c) 2001-2004 Idiap Research Institute (Ronan Collobert, Samy Bengio, Johnny Mariethoz)
+#
+# From Caffe2:
+#
+# Copyright (c) 2016-present, Facebook Inc. All rights reserved.
+#
+# All contributions by Facebook:
+# Copyright (c) 2016 Facebook Inc.
+#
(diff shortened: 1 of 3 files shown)


build_tools/collect_env.py
```diff
@@ -0,0 +1,412 @@
...
+# All other contributions:
+# Copyright(c) 2015, 2016 the respective contributors
+# All rights reserved.
+#
+# Caffe2 uses a copyright model similar to Caffe: each contributor holds
+# copyright over their contributions to Caffe2. The project versioning records
+# all such contribution and copyright details. If a contributor wants to further
+# mark their specific copyright on a particular contribution, they should
+# indicate their copyright solely in the commit message of the change when it is
+# committed.
```
dpacgopinath, line 51: What is the source of these headers, and are they are necessary considering that the [torch source file](https://github.com/pytorch/pytorch/blob/master/torch/utils/collect_env.py) does not have them?

build_tools/collect_env.py
```diff
@@ -0,0 +1,412 @@
...
+    return out.split("\n")[0]
+
+
+def get_conda_packages(run_lambda):
+    conda = os.environ.get("CONDA_EXE", "conda")
+    out = run_and_read_all(run_lambda, conda + " list")
+    if out is None:
+        return out
+    # Comment starting at beginning of line
+    comment_regex = re.compile(r"^#.*\n")
```
dpacgopinath, line 170: Would this cover comments using triple quotes? `""" Comment """`

compiler_gym/util/gym_type_hints.py
```diff
@@ -2,8 +2,8 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
 from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
```
dpacgopinath, line 6: I'd probably move this to another PR since it's an unrelated (but valid) change

dpacgopinath (approved):
This looks like a straight-forward ship-it considering there are only a few deviations from `torch/utils/collect_env.py`. I've left a few comments/questions

build_tools/collect_env.py
ChrisCummins: Hmm good question, I just inlined the entirety of PyTorch's [LICENSE](https://github.com/pytorch/pytorch/blob/master/LICENSE) file, since it uses a somewhat non-standard text

build_tools/collect_env.py
ChrisCummins: It wouldn't, but I think this is just to filter the output of `conda list`, which sometimes prefixes comment lines with `# <something>`

compiler_gym/util/gym_type_hints.py
ChrisCummins: Yeah good catch, it crept into this PR as a tiny fixup 🙂

ChrisCummins:
Many thanks for the review @dpacgopinath!

PR #692, opened 2022-05-29
Added support for loading URLs in CompilerEnvStateReader.read_paths()

### LINKED ISSUE #148: Add support for loading URLs to CompilerEnvStateReader.read_paths()
## 🚀 Feature

Extend [CompilerEnvStateReader.read_paths()](https://github.com/facebookresearch/CompilerGym/blob/de07d4867e0bb0b47f6fa4bce5e262ea8f014c3e/tests/compiler_env_state_test.py#L212-L335) so that any combination of file path or URL can be loaded.

thecoblack:
Fixes #148 Hi! This PR adds support for loading content from URLs. Regarding the tests, CompilerEnvStateReader.read_paths does not have any coverage yet, so I have added tests for other method features like input pipe or paths. LMK what you think :)!

### DIFF
--- a/compiler_gym/compiler_env_state.py
+++ b/compiler_gym/compiler_env_state.py
@@ -4,9 +4,12 @@
 # LICENSE file in the root directory of this source tree.
 """This module defines a class to represent a compiler environment state."""
 import csv
+import re
 import sys
+from io import StringIO
 from typing import Iterable, List, Optional, TextIO
 
+import requests
 from pydantic import BaseModel, Field, validator
 
 from compiler_gym.datasets.uri import BenchmarkUri
@@ -23,10 +26,7 @@ class CompilerEnvState(BaseModel):
 
     benchmark: str = Field(
         allow_mutation=False,
-        examples=[
-            "benchmark://cbench-v1/crc32",
-            "generator://csmith-v0/0",
-        ],
+        examples=["benchmark://cbench-v1/crc32", "generator://csmith-v0/0",],
     )
     """The URI of the benchmark used for this episode."""
 
@@ -37,9 +37,7 @@ class CompilerEnvState(BaseModel):
     """The walltime of the episode in seconds. Must be non-negative."""
 
     reward: Optional[float] = Field(
-        required=False,
-        default=None,
-        allow_mutation=True,
+        required=False, default=None, allow_mutation=True,
     )
     """The cumulative reward for this episode. Optional."""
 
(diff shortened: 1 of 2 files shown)


compiler_gym/compiler_env_state.py
```diff
@@ -229,6 +231,16 @@ def read_paths(paths: Iterable[str]) -> Iterable[CompilerEnvState]:
         for path in paths:
             if path == "-":
                 yield from iter(CompilerEnvStateReader(sys.stdin))
+            elif (
+                re.match(r"^(http|https)://[a-zA-Z0-9.-_/]+(\.csv)$", path) is not None
+            ):
+                response: requests.Response = requests.get(path)
```
ChrisCummins, line 233: I think you're missing an `import requests` in this module

tests/compiler_env_state_test.py
```diff
@@ -321,5 +322,81 @@ def test_state_serialize_deserialize_equality_no_reward():
...
+        CompilerEnvState(
+            benchmark="benchmark://cbench-v0/foo",
+            walltime=5,
+            commandline="-a -b -c",
+            reward=2,
+        )
+    ]
+
+
+def test_read_paths_directory(tmp_path):
```
ChrisCummins:
```suggestion
def test_read_paths_file(tmp_path):
```

Nitpick. Looks like this is testing reading from a file

tests/compiler_env_state_test.py
```diff
@@ -321,5 +322,81 @@ def test_state_serialize_deserialize_equality_no_reward():
...
+            "benchmark,reward,walltime,commandline\n"
+            "benchmark://cbench-v0/foo,2.0,5.0,-a -b -c\n"
+        )
+        status_code = 200
+
+    def get_mock_response(*args, **kwargs):
+        return MockResponse()
+
+    monkeypatch.setattr(requests, "get", get_mock_response)
+    urls = ["https://fakeurl.com/benchmarktest.csv"]
```
ChrisCummins: Minor suggestion: use a real domain for tests, and preferably one that you own. Even though the URL isn't being accessed here, `https://fakeurl.com` is a real domain and you don't necessarily want to condone its use 🙂 We own the `compilergym.ai` domain, so how about: `https://compilergym.ai/benchmarktest.csv`

ChrisCummins:
Hi @thecoblack, thanks for the patch! Your code looks clean and I appreciate the tests, thank you.

It looks like there is a missing module import. Couple of other minor suggestions left inline.

Cheers,
Chris

ChrisCummins:
LGTM, thanks @thecoblack!

Cheers,
Chris

Issue #596, opened 2022-03-04
Remove the Bazel CI steps


ChrisCummins:
## 🚀 Feature

Once ready to migrate fully from Bazel to CMake, remove the Bazel build steps from the CI jobs, replacing with their equivalent CMake versions.

PR #642, opened 2022-03-31
Enable and Fix Environment Examples in CMake CI

mostafaelhoushi:
Fixes #631

- ensured that the example scripts are invoked in CMake CI by creating `example_without_bazel_test.py` scripts
- ensured that unit tests in `env_tests.py` are invoked in CMake CI by copying them into `env_without_bazel_test.py`
- disabled the tests cases in Linux and MacOS bazel builds for now (hopefully will fix that with #707 )

PR #695, opened 2022-05-31
Run examples in cmake CI

mostafaelhoushi:
Addresses #596

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,7 +85,10 @@ jobs:
               uses: actions/upload-artifact@v2
               with:
                   name: linux-wheel-cmake
-                  path: ~/cmake_build/py_pkg/dist/*.whl
+                  path: |
+                    ~/cmake_build/py_pkg/dist/*.whl
+                    ~/example_unrolling_service/loop_unroller/loop_unroller
+                    ~/loop_optimizations_service/opt_loops/opt_loops
                   if-no-files-found: error
                   retention-days: 7
 
@@ -233,6 +236,12 @@ jobs:
               uses: actions/download-artifact@v2
               with:
                   name: linux-wheel-cmake
+                  
+            - name: Move examples binaries
+              run: |
+                  mv loop_unroller ~/example_unrolling_service/loop_unroller/loop_unroller
+                  mv opt_loops ~/loop_optimizations_service/opt_loops/opt_loops
+              shell: bash
 
             - name: Install wheel
               run: python -m pip install *.whl
@@ -245,6 +254,15 @@ jobs:
 
             - name: Run the test suite
               run: make install-test PYTEST_ARGS="--ignore tests/mlir"
+  
+            - name: Run environment examples
+              run: |
+                  cd examples
+                  python example_compiler_gym_service/demo_without_bazel.py
+                  python example_unrolling_service/example_without_bazel.py
+                  python loop_optimizations_service/example_without_bazel.py
+                  cd -
+              shell: bash
 
     test-mlir-env-linux-cmake:
         needs: build-mlir-env-linux-cmake
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -278,7 +278,7 @@ jobs:
                   mv opt_loops ./examples/loop_optimizations_service/opt_loops/opt_loops
                   echo "pwd: "
                   pwd
-                  ls -R | grep loop_unroller
+                  ls -alR | grep loop_unroller
                   test -d examples/example_unrolling_service/loop_unroller/
                   test ./examples/example_unrolling_service/loop_unroller/loop_unroller
               shell: bash
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -259,7 +259,7 @@ jobs:
 
             - name: Install test dependencies
               run: python -m pip install -r tests/requirements.txt
-              
+
             - name: Download loop_unroller
               uses: actions/download-artifact@v2
               with:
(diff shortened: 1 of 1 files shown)


ChrisCummins:
For `example_compiler_gym_service/demo_without_bazel.py`, I wrote a simple "smoke test" that runs the main entry point:

https://github.com/facebookresearch/CompilerGym/blob/2e3b1769b9409e268c1af1385201d46ad17cab3c/examples/example_compiler_gym_service/demo_without_bazel_test.py#L5-L12

I believes this should be run by the CI's example test jobs. Would this cover your needs for the other files?

Cheers,
Chris

mostafaelhoushi:
> I believes this should be run by the CI's example test jobs. Would this cover your needs for the other files?

OK. Got it. I will update the CI accordingly... 
How can we ensure that the new tests I create will only run in the cmake job? 

The point is that the other 2 tests (loop_unroller and opt_loops) require building some executables. I have tried in this CI adding the other tests: https://github.com/facebookresearch/CompilerGym/pull/642 but I found myself opening a can of worms trying to get those test files working for different jobs that use bazel, CMake, or pip install.

If I can get those examples to work with CMake only, and then look into getting them to work in other jobs that would be great.

ChrisCummins:
> The point is that the other 2 tests (loop_unroller and opt_loops) require building some executables.

Ah, sorry, I see now. 

> If I can get those examples to work with CMake only, and then look into getting them to work in other jobs that would be great.

I think you need to use CMake's testing `ctest` program to run the tests:

https://cmake.org/cmake/help/book/mastering-cmake/chapter/Testing%20With%20CMake%20and%20CTest.html

TBH I'm not super familiar with it, but you could maybe write a test job that depends on the cmake build and runs `ctest` on the examples dir?

Cheers,
Chris

mostafaelhoushi:
So far, this is working:
- running the `example_unrolling_service` and `loop_optimizations_service` python examples inside the `build-linux-cmake` job
- uploading the `loop_unroller` and `opt_loops` binaries inside the `build-linux-cmake` job
- downloading the `loop_unroller` and `opt_loops` binaries inside the `test-linux-cmake` job

However, this is not working:
- running the `example_unrolling_service` and `loop_optimizations_service` python examples inside the `test-linux-cmake` job
It returns an error saying that it can't find the `loop_unroller` binary, although I made sure it's moved to its correct directory.

mostafaelhoushi:
OK. I just noticed that about GitHub's upload/download artifcats: https://github.com/actions/download-artifact#permission-loss:
```
❗ File permissions are not maintained during artifact upload ❗ 
```

So now I have `chmod` on the binaries after downloading them. Hopefully this time it will work.

ChrisCummins:
Good progress @mostafaelhoushi!

ChrisCummins:
@mostafaelhoushi, these shouldn't be necessary:

cummins-mbp 2022-06-09 at 08 10 44

Do you see the UI button to [re-run jobs](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs)?

mostafaelhoushi:
> Do you see the UI button to [re-run jobs](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs)?

For some reason that "re-run jobs" button doesn't always appear to me. Most of the time it is not there :(

ChrisCummins:
Oh weird, but sometimes it is? So not a permission issue?

ChrisCummins:
here is what the "details" screen for the current running CI job looks like:
cummins-mbp 2022-06-09 at 08 21 44

If I were to hit Cancel, I'd have to wait a few seconds/minutes for the job to finish cancelling before the Re-try button shows up.

Cheers,
Chris

mostafaelhoushi:
> here is what the "details" screen for the current running CI job looks like: 
> If I were to hit Cancel, I'd have to wait a few seconds/minutes for the job to finish cancelling before the Re-try button shows up.

Actually, sometimes I would see the "re-run job" button in that "Details" screen and sometimes I don't.

To make things even more confusing, now when I click on the "Checks" tab of the PR, I do see the "re-run jobs" button, but earlier today I couldn't see it:
image

mostafaelhoushi:
Only one job is remaining and it says "Waiting for a runner to pick up this job..." is this expected?

image

ChrisCummins:
Hmm, guess we're stuck in a queue somewhere

ChrisCummins:
Bit just be one of those "grab a cup of coffee and come back later" messages

mostafaelhoushi:
Never mind... the test ran... the problem was solved... but now hit another error:
```
NotImplementedError: [Errno 2] No such file or directory: '/home/runner/work/CompilerGym/CompilerGym/examples/loop_optimizations_service/service_py/./../../../compiler_gym/third_party/programl/compute_programl'
```

I think that should be easy to solve: upload that `compute_programl` binary and download it.
However, I believe in the future that binary should be part of the *.whl file.

ChrisCummins:
Is there a difference between `compute_programl` and running `compute_observation Programl /path/to/bitcode.bc`?

The [compute_observation](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/ComputeObservation.cc) binary is in the wheel

mostafaelhoushi:
> Is there a difference between `compute_programl` and running `compute_observation Programl /path/to/bitcode.bc`?
> 
> The [compute_observation](https://github.com/facebookresearch/CompilerGym/blob/development/compiler_gym/envs/llvm/service/ComputeObservation.cc) binary is in the wheel

OK. I wasn't aware of `compute_observation`. I created an Issue to replace `compute_programl` with `compute_observation` : https://github.com/facebookresearch/CompilerGym/issues/702

mostafaelhoushi:
build-linux-cmake and test-linux-cmake are now green. We can merge it now.

In upcoming PRs I will remove the calls to the examples from ci.yaml and ensure that pytest runs them instead.

mostafaelhoushi:
build-linux-cmake and test-linux-cmake are now green. We can merge it now.

In upcoming PRs I will remove the calls to the examples from ci.yaml and ensure that pytest runs them instead.

Issue #566, opened 2022-02-08
CMake build gets stuck on ProGraML bazel step


ChrisCummins:
## 🐛 Bug

I have encountered a strange error where the CMake build appears to get "stuck" after the bazel build step of ProGraML. The build appears to hang indefinitely. Cancelling and re-running the CMake build works fine.

## To Reproduce

```sh
$ cmake ....... # lots of flags (see below)

# <snip>
[3/9] Performing build_programl step for 'programl'
Starting local Bazel server and connecting to it...
Loading:
Loading: 0 packages loaded
Analyzing: 12 targets (5 packages loaded, 0 targets configured)
Analyzing: 12 targets (35 packages loaded, 209 targets configured)
Analyzing: 12 targets (50 packages loaded, 3390 targets configured)
INFO: Analyzed 12 targets (50 packages loaded, 6035 targets configured).
INFO: Found 12 targets...
[3 / 157] [Prepa] BazelWorkspaceStatusAction stable-status.txt
[84 / 227] Compiling libs/exception/src/clone_current_exception_non_intrusive.cpp; 0s linux-sandbox ... (32 actions, 31 running)
[94 / 227] Compiling absl/base/internal/sysinfo.cc; 1s linux-sandbox ... (32 actions, 31 running)
[111 / 227] Compiling libs/regex/src/c_regex_traits.cpp; 2s linux-sandbox ... (32 actions, 31 running)
[115 / 227] Compiling libs/regex/src/c_regex_traits.cpp; 3s linux-sandbox ... (32 actions running)
[119 / 227] Compiling libs/regex/src/c_regex_traits.cpp; 5s linux-sandbox ... (32 actions running)
[140 / 237] Compiling libs/regex/src/c_regex_traits.cpp; 6s linux-sandbox ... (32 actions running)
[156 / 243] Compiling libs/regex/src/c_regex_traits.cpp; 8s linux-sandbox ... (32 actions, 31 running)
[174 / 250] Compiling libs/regex/src/cregex.cpp; 10s linux-sandbox ... (32 actions, 31 running)
[220 / 270] Compiling libs/regex/src/cregex.cpp; 12s linux-sandbox ... (32 actions, 31 running)
INFO: From Compiling labm8/cpp/string.cc:
external/labm8/labm8/cpp/string.cc:135:22: warning: implicit conversion from 'int' to 'char' changes value from 226 to -30 [-Wconstant-conversion]
        to.append(1, 226);
[...]

PR #697, opened 2022-06-08
timeout programl build

mostafaelhoushi:
Temporary workaround for https://github.com/facebookresearch/CompilerGym/issues/566

### DIFF
--- a/external/programl/CMakeLists.txt
+++ b/external/programl/CMakeLists.txt
@@ -33,7 +33,7 @@ externalproject_add_step(
     build_programl
     ALWAYS TRUE
     COMMAND
-        "${CMAKE_COMMAND}" -E env "CC=${CMAKE_C_COMPILER}"
+        timeout 10m "${CMAKE_COMMAND}" -E env "CC=${CMAKE_C_COMPILER}"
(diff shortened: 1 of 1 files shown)


mostafaelhoushi:
Build-linux-cmake paased but took more than 3.5 hours. So will abandon this PR.

ChrisCummins:
Interesting! So perhaps the problem is not with bazel?

mostafaelhoushi:
> Interesting! So perhaps the problem is not with bazel?

I am not sure. I will double check by reducing the timeout from 30m to like 30 seconds to check if the timeout takes effect.

On another note, in my own Linux build, it seems to hang somewhere during the build of Programl or at the end of of building Programl? See the screenshot below:
image

ChrisCummins:
> On another note, in my own Linux build, it seems to hang somewhere during the build of Programl or at the end of of building Programl? See the screenshot below:

Yeah that's the same problem as the CI is encountering, so its good that you can reproduce it locally. I believe you could probably delete the programl subdir in the `build/` directory to force a quick rebuild, rather than having to wait for the CI to do a full build

Cheers,
Chris

mostafaelhoushi:
Abandoning this PR as it seems the problem is solved in #703

PR #699, opened 2022-06-09
enable `example_compiler_gym_service` unit tests in CI

mostafaelhoushi:
Let's do things one step at a time.
In this PR I checked out one file from https://github.com/facebookresearch/CompilerGym/pull/642/files# that doesn't require external files, binaries, or benchmarks and making sure it passes in CI.


mostafaelhoushi:
@ChrisCummins : this CI seems to pass. Can we merge it? It just copied the env_tests.py unit tests to a file that can run in test-cmake-linux.

Before this PR, those test cases are not covered by CI.

mostafaelhoushi:
The CI jobs that are red are not related to this PR

ChrisCummins:
LGTM thanks @mostafaelhoushi!

PR #705, opened 2022-06-11
Jotaibench

canesche:
Adding Benchmark

Benchmark Jotai is a subset of Anghabench which is only composed of programs that have been modified to be executable.

This pull request introduces the benchmark and its dependencies.


compiler_gym/test_jotai.py
```diff
@@ -0,0 +1,36 @@
+import gym
```
ChrisCummins: I don't think this feels needs to be commited

compiler_gym/envs/llvm/datasets/jotaibench.py
```diff
@@ -0,0 +1,243 @@
...
+                    ],
+                ).command(outpath=tmp_path)
+                subprocess.check_call(compile_cmd, timeout=300)
+
+        benchmark = BenchmarkWithSource.create(
+            uri, bitcode_abspath, "function.c", c_file_abspath
+        )
+
+        # TODO: Here is where we specify how to build and run the benchmark.
+        # This is what makes a benchmark "runnable".
```
ChrisCummins: The `TODO`s in this file can be removed

compiler_gym/envs/llvm/datasets/jotaibench.py
```diff
@@ -0,0 +1,243 @@
...
+        site_data_base: Path,
+        sort_order: int = 0,
+        manifest_url: Optional[str] = None,
+        manifest_sha256: Optional[str] = None,
+        deprecated: Optional[str] = None,
+        name: Optional[str] = None,
+    ):
+        manifest_url_, manifest_sha256_ = {
+            "linux": (
+                "http://cuda.dcc.ufmg.br/Jotai/src/Jotai_printRetVal.tar.bz2",
```
ChrisCummins: Would you be happy for us to host a mirror of this file? If so, that's great, I can add it in a follow-up PR.

ChrisCummins (commented):
Hi @canesche, many thanks for this! Looking forward to integrate your dataset.

There are some test failures reported by the CI. Please fix them and run `make install install-test` to check locally that everything is as expected.

There are also some code formatting inconsistencies. You can check these (and fix most of them automatically) through pre-commit. See: https://github.com/facebookresearch/CompilerGym/blob/development/.pre-commit-config.yaml#L2-L17

Cheers,
Chris

compiler_gym/test_jotai.py
canesche: Ok, I'll remove it.

ChrisCummins:
A couple of bits of documentation need to be updated for this new dataset. Please add an entry in the table here:

https://github.com/facebookresearch/CompilerGym/blob/development/docs/source/llvm/index.rst#datasets

And then add an `.. autoclass::` entry in the list of datasets here:

https://github.com/facebookresearch/CompilerGym/blob/development/docs/source/llvm/api.rst#datasets

Both are alphabetically sorted.

Cheers,
Chris

compiler_gym/envs/llvm/datasets/jotaibench.py
pronesto: Hi Chris. I will leave that up to you. If you prefer, that's fine. But we will be updating the benchmarks, so the Jotai collection might change in the future.

ChrisCummins: Okay, sounds good, thanks 🙂

compiler_gym/envs/llvm/datasets/__init__.py
```diff
@@ -261,6 +262,26 @@ def get_llvm_datasets(site_data_base: Optional[Path] = None) -> Iterable[Dataset
...
+        ),
+    }[sys.platform]
+    yield JotaiBenchDataset(
+        name="benchmark://jotaibench-v0",
+        site_data_base=site_data_base,
+        sort_order=0,
+        manifest_url=jotaibench_v0_manifest_url,
+        manifest_sha256=jotaibench_v0_manifest_sha256,
+        deprecated="Please use jotaibench-v1",
+    )
```
ChrisCummins: Is there a particular reason to add a deprecated dataset in the same PR in which you add the replacement? What does `jotaibench-v0` do that `jotaibench-v1` does not?

canesche: No, there isn't. I'll update the code. Thanks for the observation.

ChrisCummins:
Hi @canesche, getting there, but now the version number needs downgrading 🙂 . If you do a quick find-and-replace for `jotaibench-v1` to `jotaibench-v0` then I can land this, assuming no problems with the tests

canesche:
Hi @ChrisCummins, ok I'll replace the jotaibench's version.

Issue #709, opened 2022-06-13
🔥 CI/www-build is on fire


ChrisCummins:
Issue to track the build failure of the CI/www-build job. Example error:

```
added 1827 packages, and audited 2028 packages in 1m
205 packages are looking for funding
  run `npm fund` for details
6 high severity vulnerabilities
To address all issues (including breaking changes), run:
  npm audit fix --force
Run `npm audit` for details.
> compiler_gym-frontend@0.1.0 build
> react-scripts --openssl-legacy-provider build
Creating an optimized production build...
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.
Failed to compile.
Warning
(6:[27](https://github.com/facebookresearch/CompilerGym/runs/6867197296?check_suite_focus=true#step:5:28)005) autoprefixer: Replace color-adjust to print-color-adjust. The color-adjust shorthand is currently deprecated.
```

@sahirgomez1 agreed to take a look at this later this week.

sahirgomez1:
Closing this issue with #729

PR #716, opened 2022-06-16
Add support for granular timeouts

### LINKED ISSUE #283: Add support for granular timeouts
## 🚀 Feature

The compiler environment uses timeouts on all operations. These timeouts are controlled by a [`ConnectionOpts`](https://github.com/facebookresearch/CompilerGym/blob/641e8cf6e3199b1bb7202415e02e78bf9e654f93/compiler_gym/service/connection.py#L46-L83) class, which is default-constructed by the constructor. This feature request proposes adding the ability for user code to override these timeouts on a per-method basis by passing in an `timeout` kwarg to operations that can timeout.

## Motivation

Timeout specification is currently very coarse-grained. E.g. `ConnectionOpts. rpc_call_max_seconds` specifies the timeout of all RPC operations, regardless of what the operation is, the size of the benchmark that is being processed, etc. A timeout that would be suitable for a fast operation like ending a session would be totally unsuitable for a slow operation like applying a long sequence of expensive actions on a large benchmark. Currently the only way to tweak these timeouts to match the different calls is to manually change the `env.service.opts` fields in between method calls, e.g.:

```py
>>> env = gym.make("llvm-v0", connection_settings=ConnectionOpts(rpc_call_max_seconds=10))
>>> env.reset()
>>> env.connection.opts.rpc_call_max_seconds = 60
>>> env.step(0)
>>> env.connection.opts.rpc_call_max_seconds = 10
>>> env.reset()
```

## Pitch

Add an optional `timeout` kwarg so that timeouts can be overriden on a per-method basis:

```py
>>> env = gym.make("llvm-v0")
>>> env.reset(timeout=10)
>>> env.step(0, timeout=60)
>>> env.reset(timeout=10)
```

The following methods would receive this new argument:

- [ ] `compiler_gym.envs.CompilerEnv.step`
- [ ] `compiler_gym.envs.CompilerEnv.multistep`
- [ ] `compiler_gym.envs.CompilerEnv.reset`
- [ ] `compiler_gym.envs.CompilerEnv.__init__` 

All subclasses / wrappers would need to handle this too.

## Alternatives

We could also consider ditching the [ConnectionOpts](https://github.com/facebookresearch/CompilerGym/blob/641e8cf6e3199b1bb7202415e02e78bf9e654f93/compiler_gym/service/connection.py#L46-L83) class entirely and pushing all of the configurable options out to the individual methods, but I see there are two drawbacks to this approach:

1. It makes it more difficult to support the use case where we don't need granular timeout control, e.g. where the usere would just like to specify a longer timeout that can be applied to all operations.
2. Constructor timeouts would still need to saved so that they can used in [fork()](https://github.com/facebookresearch/CompilerGym/blob/641e8cf6e3199b1bb7202415e02e78bf9e654f93/compiler_gym/envs/compiler_env.py#L539-L544) to propagate the same settings.

mar-galho:
Fixes #283.

- Added timeout kwarg to CompilerEnv and related classes.
- Added deprecation note to rpc_call_max_seconds in ConnectionOpts.


compiler_gym/envs/compiler_env.py
```diff
@@ -88,6 +88,16 @@ def close(self):
...
+    @property
+    @abstractmethod
+    def timeout(self) -> float:
+        raise NotImplementedError("abstract method")
+
+    @timeout.setter
+    @abstractmethod
+    def timeout(self, timeout: Optional[float]):
+        raise NotImplementedError("abstract method")
+
```
ChrisCummins: Is this property needed? I don't think it's used anywhere, and `timeout` arg is added as a param to the relevant methods.

compiler_gym/service/connection.py
```diff
@@ -52,6 +52,8 @@
 class ConnectionOpts(BaseModel):
     """The options used to configure a connection to a service."""
 
+    # Deprecated.
+    # This is for backwards compatibility and will be removed in a future release.
     rpc_call_max_seconds: float = 300
     """The maximum number of seconds to wait for an RPC method call to succeed."""
```
ChrisCummins: I'd suggest deleting this. We don't have any way to emit a deprecation warning to code, and now that this param is unused, it would silently break code that expected to change it. Best make that break explicit.

ChrisCummins (commented):
Hey Ricardo,

Thanks for working on this! I left a couple of comments inline. Also, it would be great if you add tests to cover this new param, perhaps by mocking the underling call to `env.service` to check that the parameter is forwarded correctly.

Please ignore the failing `CI/www` jobs, it is a known issue and can be ignore #709.

Cheers,
Chris

compiler_gym/envs/compiler_env.py
mar-galho: Done!

compiler_gym/service/connection.py
mar-galho: Done!

mar-galho:
I'll work on the tests. 

Should I add them to the `compiler_env_test.py` file only? It seems that the `connection_test.py` file is already covering the timeout kwarg, so I am assuming that, like you said, we should only check whether the kwarg is being passed through to the `CompilerGymServiceConnection` call...am I right?

ChrisCummins:
Thanks @ricardoprins!

> Should I add them to the compiler_env_test.py file only? It seems that the connection_test.py file is already covering the timeout kwarg, so I am assuming that, like you said, we should only check whether the kwarg is being passed through to the CompilerGymServiceConnection call...am I right?

Yep, that sounds like a good idea.

Cheers,
Chris

compiler_gym/service/client_service_compiler_env.py
```diff
@@ -312,6 +318,14 @@ def __init__(
...
         self.reward_space = reward_space
 
+    @property
+    def timeout(self) -> float:
+        return self._timeout
+
+    @timeout.setter
+    def timeout(self, timeout: Optional[float]):
+        self._timeout = timeout
+
```
ChrisCummins: Same here right?

compiler_gym/wrappers/core.py
```diff
@@ -126,6 +128,14 @@ def render(
...
         return self.env.render(mode)
 
+    @property
+    def timeout(self) -> float:
+        return self.env.timeout
+
+    @timeout.setter
+    def timeout(self, timeout: float):
+        self.env.timeout = timeout
+
```
ChrisCummins: and here

compiler_gym/service/client_service_compiler_env.py
mar-galho: Thanks, fixed.

compiler_gym/wrappers/core.py
mar-galho: Thanks, fixed.

ChrisCummins:
LGTM, thanks @ricardoprins!

Cheers,
Chris

PR #717, opened 2022-06-16
[bugfix] fix bug where we are not running the correct number of warmu…

lqwk:
This seems like a typo in the original code.
We are running warmup runs `getRuntimesPerObservationCount` times where we should be actually running it `getWarmupRunsPerRuntimeObservationCount` times


ChrisCummins:
Hi @lqwk, this was clearly a bug. Thanks for the patch!

Cheers,
Chris

PR #722, opened 2022-06-28
[core] Add ActionSpace.to_string(), deprecate commandline().

ChrisCummins:
This adds a new class, `compiler_gym.spaces.ActionSpace`, that wraps
the existing `env.action_space` types, and provides a pair of methods
`to_string()` and `from_string()` for serializing and deserializing
to/from string, respectively. This replaces the old
`CompilerEnv.commandline()` method, which has been marked deprecated
and will be removed in a future release.

To update your code, replace:

    env.commandline()

With:

    env.action_space.to_string(env.actions)

And replace:

    env.actions_from_commandline("...")

With:

    env.action_space.from_string("...")

The reason for this change is that "command line" is a specialized
term that does not make sense for all compiler environments. The new
ActionSpace class also provides default implementations of `to_string()`
and `from_string()`, so there is no need to implement these two
methods when providing a new backend. As a result, the LoopToolEnv
class has been removed.


dejangrubisic (approved):
Looks good to me.

ChrisCummins:
Thanks @dejangrubisic. I broke a handful of tests last week. Fixing them here, then will merge once CI is green

ChrisCummins:
Revisiting this after landing v0.2.5. I've rebased on development, will merge if/when tests pass

dejangrubisic (approved):
Looks good to me.

PR #736, opened 2022-07-25
Fix bug in llvm_rl benchmark when setting a max_benchmarks

### LINKED ISSUE #654: Bug in llvm_rl benchmark when setting a max_benchmarks
https://github.com/facebookresearch/CompilerGym/blob/b803856257aa663b20a26e49125d5d603f84a9b9/examples/llvm_rl/model/benchmarks.py#L96-L98

should be

`n = min(self.max_benchmarks, n)`

mar-galho:
Fixes #654.


ChrisCummins:
LGTM, thanks @ricardoprins!

The tests for this code are currently failing (due to an unrelated regression), so merging this under the faith that nothing is broken 🙂 

Cheers,
Chris

PR #739, opened 2022-07-26
Fix to issue #648

### LINKED ISSUE #648: Logical Failure when combing TimeLimit Wrapper with IterateOverBenchmarks
## 🐛 Bug

If an environment is first wrapper with TimeLimit before IterateOverBenchmarks, it will not return "done" as True. 

## To Reproduce

```
env = TimeLimit(env, step_limit) 
env = CycleOverBenchmarks(env, benchmarks) 
_, done, _, _ = env.reset()
while not done:
    _, done, _, _ = env.step(0) 
```
This will not finish. However, if the TimeLimit happens after the Cycle, it has normal behavior. 

## Additional context

Assign it to me, I will fix it when I got time.

nluu175:
Fixes #648

### DIFF
--- a/compiler_gym/wrappers/time_limit.py
+++ b/compiler_gym/wrappers/time_limit.py
@@ -32,18 +32,8 @@ class TimeLimit(CompilerEnvWrapper):
         self._max_episode_steps = max_episode_steps
         self._elapsed_steps = None
 
-    # def step(self, action: ActionType, **kwargs):
-    #     assert (
-    #         self._elapsed_steps is not None
-    #     ), "Cannot call env.step() before calling reset()"
-    #     observation, reward, done, info = self.env.step(action, **kwargs)
-    #     self._elapsed_steps += 1
-    #     if self._elapsed_steps >= self._max_episode_steps:
-    #         info["TimeLimit.truncated"] = not done
-    #         done = True
-    #     return observation, reward, done, info
-
     def multistep(self, actions: Iterable[ActionType], **kwargs):
+        actions = list(actions)
         assert (
             self._elapsed_steps is not None
         ), "Cannot call env.step() before calling reset()"
--- a/compiler_gym/wrappers/time_limit.py
+++ b/compiler_gym/wrappers/time_limit.py
@@ -2,7 +2,7 @@
 #
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
-from typing import Optional
+from typing import Iterable, Optional
 
 from compiler_gym.envs import CompilerEnv
 from compiler_gym.util.gym_type_hints import ActionType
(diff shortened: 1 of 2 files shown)


compiler_gym/wrappers/time_limit.py
```diff
@@ -32,12 +32,23 @@ def __init__(self, env: CompilerEnv, max_episode_steps: Optional[int] = None):
...
+    # def step(self, action: ActionType, **kwargs):
+    #     assert (
+    #         self._elapsed_steps is not None
+    #     ), "Cannot call env.step() before calling reset()"
+    #     observation, reward, done, info = self.env.step(action, **kwargs)
+    #     self._elapsed_steps += 1
+    #     if self._elapsed_steps >= self._max_episode_steps:
+    #         info["TimeLimit.truncated"] = not done
+    #         done = True
+    #     return observation, reward, done, info
```
ChrisCummins: You can delete these lines. Don't submit commented-out code

compiler_gym/wrappers/time_limit.py
```diff
@@ -32,12 +32,23 @@ def __init__(self, env: CompilerEnv, max_episode_steps: Optional[int] = None):
...
+
+    def multistep(self, actions: Iterable[ActionType], **kwargs):
```
ChrisCummins, line 35:
```suggestion
    def multistep(self, actions: Iterable[ActionType], **kwargs):
        actions = list(actions)
```

I would suggest forcing the conversion of actions to a list here, since there could be a subtle bug if actions is an iterator, as it would break the `len(actions)` below.

tests/wrappers/time_limit_wrappers_test.py
```diff
@@ -81,5 +81,21 @@ def test_time_limit_fork(env: LlvmEnv):
         fkd.close()
 
 
+# @pytest.mark.xfail(strict=True, reason="https://github.com/facebookresearch/CompilerGym/issues/648")
```
ChrisCummins: don't commit commented out code

tests/wrappers/time_limit_wrappers_test.py
```diff
@@ -81,5 +81,21 @@ def test_time_limit_fork(env: LlvmEnv):
...
+def test_time_limit(env: LlvmEnv):
+    """Check CycleOverBenchmarks does not break TimeLimit"""
+    env = TimeLimit(env, max_episode_steps=1)
```
ChrisCummins: I would suggest testing with `max_episode_steps=3`, so that you can test that `done` is False for the first two calls to `step()`.

tests/wrappers/time_limit_wrappers_test.py
```diff
@@ -81,5 +81,21 @@ def test_time_limit_fork(env: LlvmEnv):
...
+    env = CycleOverBenchmarks(
+        env,
+        benchmarks=[
+            "benchmark://cbench-v1/crc32",
+        ],
+    )
+    env.reset()
+    _, _, done, _ = env.step(0)
+
+    assert done
```
ChrisCummins, line 102:
```suggestion
    assert done
    assert info["TimeLimit.truncated"], info
```

Best to also test the reason why `done=True`.

ChrisCummins (commented):
Hi @nluu175! Thanks for the PR. I left some comments inline 🙂 

Cheers,
Chris

nluu175:
Hi @ChrisCummins . Thanks for the feedbacks. I'll resolve those and as soon as possible!

ChrisCummins:
Changes LGTM, thanks! I will merge once CI tests pass.

Cheers,
Chris

PR #740, opened 2022-07-28
Update README for llvm_autotuning

lqwk:
To specify custom log directories for `llvm_autotuning.info` I think we need to explicitly spell out the option `--log-dirs`, otherwise it doesn't work.


ChrisCummins:
LGTM, thanks!

PR #751, opened 2022-08-11
Added 2 test cases where fork failed

youweiliang:
Fork failed to produce the same rewards after applying the same action


ChrisCummins:
LGTM, thanks @youweiliang. Merging. I'll take a look at the specific problems to see if they can be fixed.

Cheers,
Chris

Issue #750, opened 2022-08-11
RL examples test is broken on CI


ChrisCummins:
## 🐛 Bug

Looks like a breaking change in a dependency. CI logs:

```
make: Entering directory '/home/runner/work/CompilerGym/CompilerGym/examples'
python3 -c 'import dgl; print(dgl.__version__)' &>/dev/null || true
pytest --no-success-flaky-report --benchmark-disable -n auto --durations=5 . --cov=compiler_gym --cov-report=xml:/home/runner/work/CompilerGym/CompilerGym/coverage.xml --ignore example_unrolling_service --ignore loop_optimizations_service
DGL backend not selected or invalid.  Assuming PyTorch for now.
Setting the default backend to "pytorch". You can change it in the ~/.dgl/config.json file or export the DGLBACKEND environment variable.  Valid options are: pytorch, mxnet, tensorflow (all lowercase)
Using backend: pytorch
0.6.1
============================= test session starts ==============================
platform linux -- Python 3.7.13, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
benchmark: 3.4.1 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/runner/work/CompilerGym/CompilerGym, configfile: tox.ini
plugins: hydra-core-1.1.0, xdist-2.4.0, timeout-1.4.2, sugar-0.9.4, stress-1.0.1, shard-0.1.2, mock-3.6.1, cov-2.12.1, benchmark-3.4.1, flaky-3.7.0, forked-1.4.0
timeout: 300.0s
timeout method: signal
timeout func_only: False
gw0 I / gw1 I
gw0 [56] / gw1 [56]

.............................ss.s.........F.............                 [100%]
=================================== FAILURES ===================================
_______________________________ test_local_train _______________________________
[gw0] linux -- Python 3.7.13 /opt/hostedtoolcache/Python/3.7.13/x64/bin/python3

Coverage.py warning: Data file '/home/runner/work/CompilerGym/CompilerGym/examples/.coverage.fv-az36-101.10042.492597' doesn't seem to be a coverage data file: Couldn't use data file '/home/runner/work/CompilerGym/CompilerGym/examples/.coverage.fv-az36-101.10042.492597': no such table: coverage_schema
[...]

PR #752, opened 2022-08-12
Add Python 3.10 support.

ChrisCummins:
This bumps several dependencies to enable Python 3.10 support, and
switches the default version of python used for single-python CI
jobs from 3.9 to 3.10.


ChrisCummins:
Fixes #750.

PR #772, opened 2022-10-31
[llvm] Add utility functions to split and merge benchmarks.

ChrisCummins:
This adds three new utility functions for constructing LLVM benchmarks:

```py
def make_benchmark_from_source(
    source: str,
    copt: Optional[List[str]] = None,
    lang: str = "c++",
    system_includes: bool = True,
    timeout: int = 600,
) -> Benchmark:
    ...

def split_benchmark_by_function(
    benchmark: Benchmark, timeout: float = 300
) -> List[Benchmark]:
    ...

def merge_benchmarks(
    benchmarks: List[Benchmark], timeout: float = 300
) -> Benchmark:
   ...
```

Credit to @hughleat for `llvm-extract-one`, an extension of LLVM's `llvm-extract` that enables functions to be extracted by an integer index rather than by name. This enables extracting anonymous functions.

## Todo before land

- [x] Merge #767 and rebase on v0.2.5 (no conflicts).
- [x] Confirm CMake + bazel builds succeed.
- [x] Add test coverage for the two new functions.

### DIFF
--- a/compiler_gym/envs/llvm/llvm_benchmark.py
+++ b/compiler_gym/envs/llvm/llvm_benchmark.py
@@ -10,6 +10,7 @@ import subprocess
 import sys
 import tempfile
 from concurrent.futures import as_completed
+from copy import deepcopy
 from datetime import datetime
 from functools import lru_cache
 from pathlib import Path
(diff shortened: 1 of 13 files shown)


compiler_gym/envs/llvm/llvm_benchmark.py
```diff
@@ -422,3 +423,80 @@ def _add_path(path: Path):
...
+                f"{stderr.decode('utf-8')}"
+            )
+    number_of_functions = int(stdout.decode("utf-8"))
+    if number_of_functions <= 0:
+        raise ValueError("No functions found!")
+
+    # Iterate over the number of functions, extracting each one in turn.
+    split_benchmarks: List[Benchmark] = []
+    for i in range(number_of_functions):
+        with Popen(
```
kevinleestone, line 575: I think you are missing `stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.PIPE`

compiler_gym/envs/llvm/llvm_benchmark.py
```diff
@@ -422,3 +423,80 @@ def _add_path(path: Path):
...
     uri = f"benchmark://user-v0/{timestamp}-{random.randrange(16**4):04x}"
     return Benchmark.from_file_contents(uri, bitcode)
+
+
+def split_benchmark_by_function(
+    benchmark: Benchmark, timeout_seconds: float = 300
+) -> List[Benchmark]:
+    """Split a benchmark into a list of benchmarks, one per function."""
+    original_uri = deepcopy(benchmark.uri)
+    original_bitcode = benchmark.proto.file.contents
```
kevinleestone: At least for me this needed to be `original_bitcode = benchmark.proto.program.contents`

compiler_gym/envs/llvm/llvm_benchmark.py
ChrisCummins: Good catch thanks!

kevinleestone (approved):
Thank you for your work on this. This LGTM.

PR #830, opened 2024-05-29
[Inst2vec] add string_of_items to fix inst2vec_preprocess.py

Panhaolin2001:
There is an error when set " observation_space='Inst2vec' ":
`NameError: name 'string_of_items' is not defined`


ChrisCummins:
LGTM, thank you for the patch @Panhaolin2001! Not sure how this slipped through.

Cheers,
Chris

PR #832, opened 2024-10-09
Bumbing http-cache-semantics to 4.1.1 to fix cve

gmatuz:
Fixes DoS issues from http-cache-semantics dependency version 4.1.0 by bumping to 4.1.1
[https://github.com/advisories/GHSA-rc47-6667-2j5j](https://github.com/advisories/GHSA-rc47-6667-2j5j)

Most of the dependencies are quite outdated but currently runs under node 13 with `npm ci` I tried to change as little as necessary

Does not work on more recent node version since py-gyp won't build probaby because the py-gyp is not maintained anymore

### DIFF
--- a/www/frontends/compiler_gym/package-lock.json
+++ b/www/frontends/compiler_gym/package-lock.json
@@ -8775,9 +8775,9 @@
       }
     },
     "node_modules/http-cache-semantics": {
-      "version": "4.1.0",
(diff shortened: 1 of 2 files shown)


ChrisCummins (approved):
LGTM, thank you for the contribution @gmatuz!

Cheers,
Chris

